PR #13, opened 2017-03-09
Fix the package name in order to publish to npm

lukeapage:


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@saxo/openapi-client",
+  "name": "openapi-clientlib",
   "version": "6.0.0",
```
lukeapage: change the version number too

PR #14, opened 2017-03-09
Bump grunt-contrib-jasmine version to fix issue with missing _.sprintf

michax:
Version of grunt-contrib-jasmine that we use had issue with printing exceptions due to missing _.sprintf:
https://github.com/gruntjs/grunt-contrib-jasmine/issues/232

I've bumped version to 1.0.3 which has fix for this issue and doesn't use deprecated solution (1.0.2 doesn't).

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -93,9 +93,9 @@
       "dev": true
     },
     "aws4": {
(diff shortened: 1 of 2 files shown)


lukeapage:
Looks good to me.

PR #15, opened 2017-03-09
Added subscription actions queue which fixes issues with dropped subscription actions.

michax:
Added subscription actions queue which fixes issues with dropped subscription actions during subscription transition.

Moved subscription actions to separate file, so subscription-queue will not depend on subscription (no cyclical dependency).

Example case when we drop subscribe actions:
```
subscribe -> unsubscribe -> subscribe
```
If first subscribe will take a long time, unsubscribe action could be missed. 
Afterwards, seconds subscribe will be dropped as subscribe will be already in subscribe state.

Issue is easily reproducible if you do quick 3 resubscribe (subscribe/unsubscribe) actions on slow network connection.

New solution was stress tested against fast and slow connections (GPRS throttling in chrome).
Using scenarios with small and big number of consecutive resubscribe calls (100).


src/openapi/streaming/subscription-queue.js
```diff
@@ -0,0 +1,88 @@
...
+ * Add pending action to a queue and making sure that we remove unecessary actions like:
+ * - duplicates.
+ * - ACTION_SUBSCRIBE action before ACTION_UNSUBSCRIBE.
+ *
+ * @param {Number} action - action to add to the queue.
+ */
+SubscriptionQueue.prototype.enqueue = function(action) {
+    this.queue.push(action);
+
+    console.log('[enque] add action: ', action, ' MaxSize: ', this.maxSize);
```
lukeapage: This should log, not console log. Does it pass linting?

src/openapi-package.js
```diff
@@ -38,6 +40,8 @@ export default {
 		// privates exposed for testing
 		_StreamingOrphanFinder: StreamingOrphanFinder,
 		_StreamingSubscription: StreamingSubscription,
+        _StreamingSubscriptionQueue: StreamingSubscriptionQueue,
```
lukeapage: There are a few indentation differences. This repo uses tabs..

test/unit/openapi/streaming/subscription-spec.js
```diff
@@ -387,7 +387,10 @@ describe("openapi StreamingSubscription", () => {
 			});
 		});
 
-		it ("ignores a unsubscribe followed by a subscribe when waiting for an action to respond", (done) => {
+		/**
+		 * Unsubscribe before subscribe is required for modify action.
+ 		 */
+		it ("accept unsubscribe followed by a subscribe when waiting for an action to respond", (done) => {
```
lukeapage: I'd rather have an explicit modify action and ignore subscribe followed by unsubscribe..

src/openapi/streaming/subscription-queue.js
michax: Leftover after debugging session. Lint didn't catch it. Removed.

src/openapi-package.js
michax: Somehow i inserted those spaces manually. Fixed.

src/openapi/streaming/subscription-queue.js
michax: Added jshint rule to disallow console usage. Also removed existing usage as it was found int print-task.

Created separate PR for it as this one will probably take some time.
https://github.com/SaxoBank/openapi-clientlib-js/pull/19

src/openapi/streaming/subscription.js
```diff
@@ -331,7 +346,7 @@ Subscription.prototype.reset = function() {
 			return;
 	}
 
-	this.nextAction = null;
+    this.queue.reset();
```
lukeapage: Another indentation problem..
We should convert to spaces and add eslint eventually but let's hold off till this is in..

michax: Done.

Tried to add jshint indentation rule but they removed it, as author decided that style checking is responsibility of another tool.

src/openapi/streaming/subscription-queue.js
```diff
@@ -0,0 +1,86 @@
...
+	ACTION_UNSUBSCRIBE
+} from './subscription-actions';
+
+/**
+ * Queue (FIFO) for storing pending subscription actions.
+ * @param maxSize {Number} - Maximum queue size. Defaults to 2, which works best for current needs.
+ * @constructor
+ */
+function SubscriptionQueue(maxSize = 2) {
+	this.queue = [];
```
ap3rus: I'd rename this to `items` or `array`, it's not a queue, but the wrapping class is

test/unit/openapi/streaming/subscription-queue-spec.js
```diff
@@ -0,0 +1,217 @@
...
+
+	describe("enqueue", () => {
+        beforeEach(() => {
+            queue = new SubscriptionQueue();
+        })
+
+		it('should add one action', () => {
+			queue.enqueue(SubscriptionActions.ACTION_SUBSCRIBE);
+			expect(queue.peek()).toBe(SubscriptionActions.ACTION_SUBSCRIBE);
+            expect(queue.isEmpty()).toBe(false);
```
ap3rus: formatting - space needed before `expect(...`

src/openapi/streaming/subscription-queue.js
michax: Done.

test/unit/openapi/streaming/subscription-queue-spec.js
michax: Done.

src/openapi/streaming/subscription-queue.js
```diff
@@ -0,0 +1,86 @@
...
+	if (this.items.length < this.maxSize) {
+		return;
+	}
+
+	for (let i = this.items.length - 2; i >= 0; i--) {
+		let task = this.items[i];
+		let nextTask = this.items[i + 1];
+
+		if (task === ACTION_SUBSCRIBE && nextTask === ACTION_UNSUBSCRIBE || task === nextTask) {
+			this.items.splice(i, 1);
```
lukeapage: What happens when you enqueue a unsubscribe.. it won't remove the initial subscribe will it?

michax: It will. Do we need initial subscribe, when it's surpassed by unsubscribe? 
Test case:
https://github.com/michax/openapi-clientlib-js/blob/ff4ec8cd774920bf5da19bc02931eaabdc1d640c/test/unit/openapi/streaming/subscription-queue-spec.js#L37

michax: Pushed one more test, showcasing same behaviour as above but against multiple actions:
https://github.com/michax/openapi-clientlib-js/blob/bug-missing-subscription-calls/test/unit/openapi/streaming/subscription-queue-spec.js#L60

michax: Or maybe i don't understand something. From what i get, subscription.js is class instantiated per subscription config. I don't see any scenario where keeping old subscription call would make sense. 

Open to hearing some scenarios that will fail with my implementation.

test/unit/openapi/streaming/subscription-spec.js
michax: Did you mean unsubscribe followed by subscribe? Cause it's what you linked to.

Anyhow, digesting that idea. I know we talked about it in December/January. But I'm not sure if it's worth complicating subscription.

michax:
I've added modify action, which now dictates that we shouldn't drop unsubscribe surpassed by subscribe. 
Now pending modify action will cause dropped pending unsubscribe in above case (update related test).

michax:
I need to add more test for cases where we call modify_subscribe surpassed by subscribe.

src/openapi/streaming/streaming.js
```diff
@@ -463,13 +463,16 @@ Streaming.prototype.createSubscription = function(serviceGroup, url, subscriptio
...
- * Makes a subscription start
+ * Makes a subscription start.
  *
  * @param {saxo.openapi.StreamingSubscription} subscription - The subscription to start.
+ * @param {Boolean} modify - The flag indicating if subscription is an modification. 
+ *                           If true, subscription will keep pending unsubscribe followed by subscribe with modify flag set to true. 
+ *                           Oherwise, unsubscribe in such scenario will be skipped.
  */
-Streaming.prototype.subscribe = function(subscription) {
+Streaming.prototype.subscribe = function(subscription, modify) {
```
lukeapage: I would rather an explicit modify method that took the new arguments.
In future this could take an option of whether to a) wait for unsubscribe b) use "patch"

michax: Yeah, I initially did that but reverted back to having 'less' changes in streaming. 
Done.

src/openapi/streaming/streaming.js
```diff
@@ -473,6 +473,17 @@ Streaming.prototype.subscribe = function(subscription) {
...
 
 /**
+ * Makes a subscription start with modification.
+ * Modify subscription will keep pending unsubscribe followed by modify subscribe.
+ *
+ * @param {saxo.openapi.StreamingSubscription} subscription - The subscription to modify.
+ */
+Streaming.prototype.modify = function(subscription) {
+
+	subscription.onSubscribe(true);
```
lukeapage: As per previous comment, I was expecting this to look like
```
modify(subscription, args) {

    subscription.onModify(args);
}
```
later an optional argument "options"
and
```
onModify(args) {
   this.onUnsubscribe();
   this.args = args;
    this.onSubscribe(true);
}
```

reasoning - otherwise we are fiddling with the classes own args (which it uses itself to auto subscribe), so I want to hand the whole modify process (including later patch option) to the library.

michax: Makes sense.

michax: Done and dev tested.

PR #23, opened 2017-03-10
use asap to share next tick rather than our own version

lukeapage:


jumoel:
I assume this is not valid anymore?

lukeapage:
I think its still worth doing, its just small benefit, so I never got around to following through

PR #24, opened 2017-03-13
Add formatting scientific notation's values support

pedroalves0:
Adds the possibility to format values on scientific notation.

A value such as `0.0000005` is automatically converted to `5e-7` on javascript. Number format would then pick convert the value to a string and the output would be something completely wrong - i.e `5,e-7`.

This fix parses the scientific notation value and adds the correct amount of zeros to then allow the formatter to do his normal steps.

### DIFF
--- a/src/number-formatting/format.js
+++ b/src/number-formatting/format.js
@@ -12,6 +12,21 @@ function formatNegativeNumber(str, options) {
 }
 
 /**
+ * converts a number to a decimal string if it is on scientific notation
+ * @param number
+ */
+function convertNumbertToString(number, precision) {
+	let numberString = String(number);
+
+	// if the number is in scientific notation, convert to decimal
+	if(/\d+\.?\d*e[\+\-]*\d+/i.test(numberString)) {
+		numberString = number.toFixed(precision).trim('0');
+	}
+
+	return numberString;
+}
+
+/**
  * expands the number of decimals and introduces decimal groups.
  * @param number
  * @param precision
@@ -22,7 +37,7 @@ function formatNegativeNumber(str, options) {
 function expandNumber(number, precision, groupSizes, sep, decimalChar) {
 	var curSize = groupSizes[0],
 			curGroupIndex = 1,
-			numberString = String(number),
+			numberString = convertNumbertToString(number, precision),
 			decimalIndex = numberString.indexOf('.'),
 			right = "",
 			i;
--- a/test/unit/number-formatting/format-spec.js
+++ b/test/unit/number-formatting/format-spec.js
@@ -24,18 +24,22 @@ describe("NumberFormatting format", () => {
 			expect(formatNumberNoRounding(0, 4)).toEqual("0.0000");
 			expect(formatNumberNoRounding(1.1, 4)).toEqual("1.1000");
 			expect(formatNumberNoRounding(1.1212, 1)).toEqual("1.1212");
+			expect(formatNumberNoRounding(5e-7, 1)).toEqual("0.0000005");
 
 			expect(formatNumberNoRounding(-1.1212, 1)).toEqual("-1.1212");
 			expect(formatNumberNoRounding(-1.1, 4)).toEqual("-1.1000");
+			expect(formatNumberNoRounding(-5e-7, 1)).toEqual("-0.0000005");
 
 			expect(formatNumberNoRounding(0, 1, 2)).toEqual("0.0");
 			expect(formatNumberNoRounding(0, 4, 6)).toEqual("0.0000");
 			expect(formatNumberNoRounding(1.1, 4, 6)).toEqual("1.1000");
 			expect(formatNumberNoRounding(1.1212, 1, 3)).toEqual("1.121");
+			expect(formatNumberNoRounding(5e-7, 4, 8)).toEqual("0.0000005");
 
 			expect(formatNumberNoRounding(-1.1212, 1, 2)).toEqual("-1.12");
 			expect(formatNumberNoRounding(-1.126, 1, 2)).toEqual("-1.13");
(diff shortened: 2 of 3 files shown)


src/number-formatting/format.js
```diff
@@ -22,7 +53,8 @@ function formatNegativeNumber(str, options) {
 function expandNumber(number, precision, groupSizes, sep, decimalChar) {
 	var curSize = groupSizes[0],
 			curGroupIndex = 1,
-			numberString = String(number),
+			decimalNumber = convertScientificToDecimal(number),
+			numberString = String(decimalNumber),
```
lukeapage: wouldn't it be clearer to rename convertScientificToDecimal to numberToString() and use it directly to get numberString ?

src/number-formatting/format.js
```diff
@@ -12,6 +12,37 @@ function formatNegativeNumber(str, options) {
...
+        var coefficient = parts[0].split('.');
+
+        if (sign === -1) {
+            number = zero + '.' + new Array(zerosLength).join(zero) + coefficient.join('');
+        } else {
+            var coefficientDecimal = coefficient[1];
+
+            if (coefficientDecimal) {
+				zerosLength = zerosLength - coefficientDecimal.length;
+			}
```
lukeapage: Sorry, but until we fix this, we have tab indentation and no linting to enforce/autofix. Please look at indentation and fix..

src/number-formatting/format.js
```diff
@@ -12,6 +12,37 @@ function formatNegativeNumber(str, options) {
 }
 
 /**
+ * converts a number to a decimal string if it is on scientific notation
+ * (method taken from https://gist.github.com/jiggzson/b5f489af9ad931e3d186)
+ * @param number
+ */
+function convertScientificToDecimal(number) {
+    // if the number is in scientific notation remove it
+    if(/\d+\.?\d*e[\+\-]*\d+/i.test(number)) {
```
lukeapage: is number coerced to a string? I'd prefer forcing it to a string at the top so it is clear (see other comment)

src/number-formatting/format.js
```diff
@@ -12,6 +12,37 @@ function formatNegativeNumber(str, options) {
...
+        var zero = '0';
+        var parts = String(number).toLowerCase().split('e');
+        var exponential = parts.pop();
```
lukeapage: here you use parts.pop and below parts[0] - maybe parts[0] and parts[1] might be easier to follow?

src/number-formatting/format.js
```diff
@@ -12,6 +12,37 @@ function formatNegativeNumber(str, options) {
...
+        if (sign === -1) {
+            number = zero + '.' + new Array(zerosLength).join(zero) + coefficient.join('');
+        } else {
+            var coefficientDecimal = coefficient[1];
+
+            if (coefficientDecimal) {
+				zerosLength = zerosLength - coefficientDecimal.length;
+			}
+
+            number = coefficient.join('') + new Array(zerosLength + 1).join(zero);
```
lukeapage: is there no helper function for `new Array(zerosLength + 1).join(zero)`

src/number-formatting/format.js
```diff
@@ -12,6 +12,37 @@ function formatNegativeNumber(str, options) {
 }
 
 /**
+ * converts a number to a decimal string if it is on scientific notation
+ * (method taken from https://gist.github.com/jiggzson/b5f489af9ad931e3d186)
```
lukeapage: lets not reference this gist since we are making changes to make it more readable (not sure what a link adds?)

test/unit/number-formatting/format-spec.js
```diff
@@ -24,18 +24,22 @@ describe("NumberFormatting format", () => {
 			expect(formatNumberNoRounding(0, 4)).toEqual("0.0000");
 			expect(formatNumberNoRounding(1.1, 4)).toEqual("1.1000");
 			expect(formatNumberNoRounding(1.1212, 1)).toEqual("1.1212");
+			expect(formatNumberNoRounding(5e-7, 1)).toEqual("0.0000005");
```
lukeapage: did you check coverage didn't go down?

test/unit/number-formatting/format-spec.js
```diff
@@ -24,18 +24,22 @@ describe("NumberFormatting format", () => {
...
 
 			expect(formatNumberNoRounding(-1.1212, 1)).toEqual("-1.1212");
 			expect(formatNumberNoRounding(-1.1, 4)).toEqual("-1.1000");
+			expect(formatNumberNoRounding(-5e-7, 1)).toEqual("-0.0000005");
```
lukeapage: add a test for a large exponent to check it works? (or are we ignoring that side?)

src/number-formatting/format.js
```diff
@@ -12,6 +12,37 @@ function formatNegativeNumber(str, options) {
 }
 
 /**
+ * converts a number to a decimal string if it is on scientific notation
```
lukeapage: also jshint failed...

src/number-formatting/format.js
pedroalves0: I'm afraid not, unless that `lodash` was being used, where a `_.repeat` helper method exists.

test/unit/number-formatting/format-spec.js
pedroalves0: I ignored that side because it takes a big value to make it be converted so scientific notation. Something like `5e21`.

I can add some asserts just for the sake of it, but for price formatting reasons I don't think it is relevant.

pedroalves0:
Pushed a new commit that simplifies the process, using `toFixed` to convert the scientific notation value to a decimal in a string (the right padding zeros are trimmed). The given precision is passed to the method.

PR #27, opened 2017-03-23
Retry creating a subscription on gateway timeouts

### LINKED ISSUE #26: Retry subscriptions on receiving 504 response
We should retry (up to 5 times?) subscribing after a few seconds if the server responds with a gateway timeout (504).

joelwilliamsmith:
If the server responds to the subscribe request with a 504 (gateway
timeout) then we need to retry creating the subscription. Retries up to 5
times every 1 second then gives up and goes to error state.

Fixes #26

### DIFF
--- a/src/openapi/streaming/subscription.js
+++ b/src/openapi/streaming/subscription.js
@@ -31,6 +31,9 @@ const MIN_REFRESH_RATE_MS = 100;
 
 const LOG_AREA = 'Subscription';
 
+const RETRY_DELAY = 1000; // retry subscribing on 504's
+const MAX_RETRIES = 5;
+
 // -- Local methods section --
 
 /**
@@ -205,7 +208,14 @@ function onSubscribeSuccess(referenceId, result) {
     }
     this.updatesBeforeSubscribed = null;
 
+    this.retryCounter = 0;
+
     onReadyToPerformNextAction.call(this);
+}
+
+function retrySubscription() {
+    this.onUnsubscribe();
+    setTimeout(this.onSubscribe.bind(this), RETRY_DELAY);
 }
 
 /**
@@ -218,13 +228,20 @@ function onSubscribeError(referenceId, response) {
         return;
     }
 
-    this.currentState = STATE_UNSUBSCRIBED;
-    log.error(LOG_AREA, 'An error occurred subscribing', { response, url: this.url });
+    if (response && response.status === '504' && this.retryCounter++ < MAX_RETRIES) {
+        log.error(LOG_AREA, 'Received gateway timeout when subscribing. Retrying.', { response, url: this.url });
 
-    // if we are unsubscribed, do not fire the error handler
-    if (this.queue.peek() !== ACTION_UNSUBSCRIBE) {
-        if (this.onError) {
-            this.onError(response);
+        this.currentState = STATE_SUBSCRIBED; // set to subscribed so that unsubscribe request isn't ignored
+        retrySubscription.call(this);
+    } else {
+        this.currentState = STATE_UNSUBSCRIBED;
+        log.error(LOG_AREA, 'An error occurred subscribing', { response, url: this.url });
+
+        // if we are unsubscribed, do not fire the error handler
+        if (this.queue.peek() !== ACTION_UNSUBSCRIBE) {
+            if (this.onError) {
+                this.onError(response);
+            }
         }
     }
     onReadyToPerformNextAction.call(this);
@@ -293,6 +310,11 @@ function Subscription(streamingContextId, transport, serviceGroup, url, subscrip
      * @type {string}
      */
     this.referenceId = null;
+
+    /**
+     * The number of retries this subscription has made
+     */
+    this.retryCounter = 0;
 
     /**
      * The action queue
--- a/test/unit/openapi/streaming/subscription-spec.js
+++ b/test/unit/openapi/streaming/subscription-spec.js
@@ -112,6 +112,52 @@ describe("openapi StreamingSubscription", () => {
                 done();
             });
         });
+
+        it("retries subscribing if receive gateway timeout", (done) => {
+            var subscription = new Subscription('123', transport, 'serviceGroup', 'test/resource', {}, createdSpy, null, errorSpy);
+            subscription.onSubscribe();
+
+            transport.postReject({ status: "504", response: { message: "Gateway Timeout"}});
+
+            tick(() => {
+                expect(transport.delete.calls.count()).toEqual(1);
+                transport.deleteResolve();
+                tick(() => {
+
+                    jasmine.clock().tick(1000);
+                    // second time round server responds with 200
+                    sendInitialResponse();
+
+                    tick(() => {
+                        expect(errorSpy).not.toHaveBeenCalled();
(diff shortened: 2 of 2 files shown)


src/openapi/streaming/subscription.js
```diff
@@ -218,13 +226,20 @@ function onSubscribeError(referenceId, response) {
...
-    this.currentState = STATE_UNSUBSCRIBED;
-    log.error(LOG_AREA, 'An error occurred subscribing', { response, url: this.url });
+    if (response.status === '504' && this.retryCounter++ < MAX_RETRIES) {
+        log.error(LOG_AREA, 'Received gateway timeout when subscribing. Retrying.', { response, url: this.url });
 
-    // if we are unsubscribed, do not fire the error handler
-    if (this.queue.peek() !== ACTION_UNSUBSCRIBE) {
-        if (this.onError) {
-            this.onError(response);
+        this.currentState = STATE_SUBSCRIBED; // assume we have an active (but faulty) subscription, so unsubscribe and try again
```
lukeapage: Comment this line that this is so unsubscribe isn't ignored?

src/openapi/streaming/subscription.js
```diff
@@ -208,6 +211,11 @@ function onSubscribeSuccess(referenceId, result) {
     onReadyToPerformNextAction.call(this);
 }
 
+function retrySubscription() {
+    this.onUnsubscribe();
+    setTimeout(this.onSubscribe.bind(this), RETRY_DELAY);
```
lukeapage: Further checks in set timeout? Ex.g. What if we unsubscribed during timeout?

src/openapi/streaming/subscription.js
```diff
@@ -295,6 +310,11 @@ function Subscription(streamingContextId, transport, serviceGroup, url, subscrip
     this.referenceId = null;
 
     /**
+     * The number of retries this subscription has made
+     */
+    this.retryCounter = 0;
```
lukeapage: Reset counter? Or I missed that?

src/openapi/streaming/subscription.js
```diff
@@ -218,13 +226,20 @@ function onSubscribeError(referenceId, response) {
         return;
     }
 
-    this.currentState = STATE_UNSUBSCRIBED;
-    log.error(LOG_AREA, 'An error occurred subscribing', { response, url: this.url });
+    if (response.status === '504' && this.retryCounter++ < MAX_RETRIES) {
```
lukeapage: Do we previously assume response is an object or non null?

src/openapi/streaming/subscription.js
joelwilliamsmith: Resetting in `onSubscribeSuccess`

michax:
Suggestion to consider. Wouldn't it make more sense to increase delay? For example, start with 1000 ms delay, and end with 5000 ms for last retry.  We could use very natural Fibonacci sequence and use 1s, 1s, 2s, 3s, 5s. Giving us total of 12s for getting back on track. What do you think?

src/openapi/streaming/subscription.js
michax: You should also reset it in subscription.reset method, and clear ongoing retry timeout if present.

I think we should also dispose of retries whenever new subscription request comes in. Especially when user requests unsubscribe.

src/openapi/streaming/subscription.js
```diff
@@ -218,13 +228,20 @@ function onSubscribeError(referenceId, response) {
...
-    this.currentState = STATE_UNSUBSCRIBED;
-    log.error(LOG_AREA, 'An error occurred subscribing', { response, url: this.url });
+    if (response && response.status === '504' && this.retryCounter++ < MAX_RETRIES) {
+        log.error(LOG_AREA, 'Received gateway timeout when subscribing. Retrying.', { response, url: this.url });
 
-    // if we are unsubscribed, do not fire the error handler
-    if (this.queue.peek() !== ACTION_UNSUBSCRIBE) {
-        if (this.onError) {
-            this.onError(response);
+        this.currentState = STATE_SUBSCRIBED; // set to subscribed so that unsubscribe request isn't ignored
```
michax: Ideally I think we should peek next action, and check if it's unsubscribe. If so, no need for retrying.

src/openapi/streaming/subscription.js
```diff
@@ -218,13 +228,20 @@ function onSubscribeError(referenceId, response) {
...
+        retrySubscription.call(this);
```
michax: It looks like we will run two actions in such case. First, retrySubscription will run unsubscribe immediately. And at the end of this function we are calling:  
```
```
Which will run next action from a queue. I don't think it's quite right.

michax: It looks like we will run two actions in such case. First, retrySubscription will run unsubscribe immediately. And at the end of this function we are calling:  
```
```
Which will run next action from a queue. I don't think it's quite right.

test/unit/openapi/streaming/subscription-spec.js
```diff
@@ -112,6 +112,52 @@ describe("openapi StreamingSubscription", () => {
...
+                    tick(() => {
+                        expect(errorSpy).not.toHaveBeenCalled();
+                        expect(createdSpy.calls.count()).toEqual(1);
+                        done();
+                    });
+                });
+            });
+        });
+
+        it("retries subscribing 5 times before giving up", (done) => {
```
michax: I think we need more tests for edge cases. 
* Unsubscribe request in the middle of retry.
* Modify request in the middle of retry.
* Subscribe request in the middle of retry. 
* Dispose (reset) request in the middle of retry.
* Modify which has timing out unsubscribe. But needs to follow up with subscribe afterwards (from queue).

src/openapi/streaming/subscription.js
```diff
@@ -205,9 +208,16 @@ function onSubscribeSuccess(referenceId, result) {
     }
     this.updatesBeforeSubscribed = null;
 
+    this.retryCounter = 0;
+
     onReadyToPerformNextAction.call(this);
 }
 
+function retrySubscription() {
+    this.onUnsubscribe();
```
michax: If Gateway timetous on subscribe request. Why do we need to unsubscribe? 
Also, what will happend if onUnsubscribeError returns timeout as well?

src/openapi/streaming/subscription.js
michax: What do you think about maybe including STATE_SUBSCRIBE_TIMEOUT? Seems a little confusing to override this.currentState with non factual one.

src/openapi/streaming/subscription.js
michax: If we indeed need to call unsubscribe first to subscribe after subscribe timeout. We should first peek into next action. If next action is unsubscribe, no need to try to retry subscription. 
```
this.queue.peek()
```

src/openapi/streaming/subscription.js
joelwilliamsmith: It could be that a proxy timed out, and the subscription was actually successful. I _assume_ the server copes with receiving an unsubscribe request for a non-existent subscription if that's not indeed the case.

lukeapage:
superseded by #29

PR #28, opened 2017-03-27
 #749180. Delete on reset, modify patch.

AlexeyKropotin:
1. Unsubscribe on reset.
2. Patch on modify with options

### DIFF
--- a/src/openapi/streaming/subscription-queue.js
+++ b/src/openapi/streaming/subscription-queue.js
@@ -14,6 +14,13 @@ import {
     ACTION_MODIFY_PATCH,
 } from './subscription-actions';
 
+function getLastItem() {
+    if (this.isEmpty()) {
+        return undefined;
+    }
+    return this.items[this.items.length - 1];
+}
+
 /**
  * Queue (FIFO) for storing pending subscription actions.
  * @param maxSize {Number} - Maximum queue size. Defaults to 2, which works best for current needs.
@@ -21,13 +28,6 @@ import {
  */
 function SubscriptionQueue(maxSize) {
     this.items = [];
-
-    this.getLastItem = function() {
-        if (this.isEmpty()) {
-            return undefined;
-        }
-        return this.items[this.items.length - 1];
-    };
 }
 
 /**
@@ -48,7 +48,7 @@ SubscriptionQueue.prototype.enqueue = function(queuedItem) {
     }
 
     const { action } = queuedItem;
-    const prevAction = this.getLastItem().action;
+    const prevAction = getLastItem.call(this).action;
 
     // ..UM => UM
     if (action === ACTION_MODIFY_SUBSCRIBE) {
@@ -105,7 +105,7 @@ SubscriptionQueue.prototype.dequeue = function() {
         return nextItem;
     }
     const nextAction = nextItem.action;
-    const lastItem = this.getLastItem();
+    const lastItem = getLastItem.call(this);
     const lastAction = lastItem.action;
 
     if (nextAction === ACTION_MODIFY_SUBSCRIBE && lastAction !== ACTION_UNSUBSCRIBE) {
--- a/src/openapi/streaming/subscription-queue.js
+++ b/src/openapi/streaming/subscription-queue.js
@@ -29,35 +29,64 @@ function SubscriptionQueue(maxSize) {
  * - duplicates.
  * - ACTION_SUBSCRIBE action before ACTION_UNSUBSCRIBE.
  *
- * @param {Number} action - action to add to the queue.
+ * @param {Object} queuedAction - action with arguments to add to the queue.
  */
-SubscriptionQueue.prototype.enqueue = function(action) {
-    this.items.push(action);
+SubscriptionQueue.prototype.enqueue = function(queuedAction) {
 
-    if (this.items.length < this.maxSize) {
+    if (!queuedAction.action) {
+        throw new Error('Subscription queued action is invalid');
+    }
+    const { action } = queuedAction;
+
+    if (!this.canEnqueueAction(action)) {
         return;
     }
 
-    for (let i = this.items.length - 2; i >= 0; i--) {
-        const task = this.items[i];
-        const nextTask = this.items[i + 1];
-
-        if (task === ACTION_MODIFY_SUBSCRIBE && nextTask === ACTION_SUBSCRIBE) {
-            // Removing subscribe in such case to keep requested subscription modification.
-            this.items.splice(i + 1, 1);
-        } else if (task === ACTION_SUBSCRIBE && nextTask === ACTION_UNSUBSCRIBE ||
-            task === ACTION_UNSUBSCRIBE && nextTask === ACTION_SUBSCRIBE ||
-            task === ACTION_SUBSCRIBE && nextTask === ACTION_MODIFY_SUBSCRIBE ||
-            task === ACTION_MODIFY_SUBSCRIBE && nextTask === ACTION_UNSUBSCRIBE ||
-            task === nextTask
-        ) {
-            this.items.splice(i, 1);
-        }
+    if (action === ACTION_UNSUBSCRIBE ||
+        action === ACTION_SUBSCRIBE) {
+        this.reset();
+    } else if (action === ACTION_MODIFY_SUBSCRIBE) {
+        this.removeAllExcept(ACTION_UNSUBSCRIBE);
     }
+
+    this.items.push(queuedAction);
 
     if (this.items.length > this.maxSize) {
         // Removes elements from the beginning of a queue, to match maximum allowed size.
         this.items.splice(0, this.items.length - this.maxSize);
+    }
+};
+
+SubscriptionQueue.prototype.canEnqueueAction = function canEnqueueAction(nextAction) {
+
+    // Removing subscribe in such case to keep requested subscription modification.
+    if (this.containsAction(ACTION_MODIFY_SUBSCRIBE) && nextAction === ACTION_SUBSCRIBE) {
+        return false;
+    }
+    return true;
+};
+
+SubscriptionQueue.prototype.containsAction = function(action) {
+    let i = 0;
+
+    while (i < this.items.length) {
+        if (this.items[i].action === action) {
+            return true;
+        }
+        i++;
+    }
+    return false;
+};
+
+SubscriptionQueue.prototype.removeAllExcept = function(action) {
+    let i = 0;
+
+    while (i < this.items.length) {
+        if (this.items[i].action !== action) {
+            this.items.splice(i, 1);
+            continue;
+        }
+        i++;
     }
 };
 
--- a/src/openapi/streaming/subscription-queue.js
+++ b/src/openapi/streaming/subscription-queue.js
@@ -7,7 +7,6 @@
  * @module saxo/openapi/streaming/subscription queue
  * @ignore
  */
-
 import {
     ACTION_SUBSCRIBE,
     ACTION_UNSUBSCRIBE,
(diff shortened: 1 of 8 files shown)


src/openapi/streaming/subscription.js
```diff
@@ -263,6 +290,32 @@ function onUnsubscribeError(referenceId, response) {
 }
 
 /**
+ * Called after modify patch is successful
+ * @param referenceId
+ * @param response
+ */
+function onModifyPatchSuccess(referenceId, response) {
+    if (referenceId !== this.referenceId) {
+        log.debug(LOG_AREA, 'Received an error response for patching a subscription');
```
lukeapage: this is not an error response..
I think its just to return because it means the subscription has reset or been unsubscribed since the patch occurred..

src/openapi/streaming/subscription.js
```diff
@@ -263,6 +290,32 @@ function onUnsubscribeError(referenceId, response) {
...
+ * Called after modify patch is successful
+ * @param referenceId
+ * @param response
+ */
+function onModifyPatchSuccess(referenceId, response) {
+    if (referenceId !== this.referenceId) {
+        log.debug(LOG_AREA, 'Received an error response for patching a subscription');
+        return;
+    }
+    onReadyToPerformNextAction.call(this);
```
lukeapage: you do not need to do this because you do not change state into something pending, so you are not releasing anything

src/openapi/streaming/subscription.js
```diff
@@ -263,6 +290,32 @@ function onUnsubscribeError(referenceId, response) {
...
+}
+
+/**
+ * Called when a unsubscribe errors
+ * @param response
+ */
+function onModifyPatchError(referenceId, response) {
+    if (referenceId !== this.referenceId) {
```
lukeapage: I would get rid of the if completely

src/openapi/streaming/subscription.js
```diff
@@ -263,6 +290,32 @@ function onUnsubscribeError(referenceId, response) {
...
+        log.debug(LOG_AREA, 'Received an error response for patching a subscription');
+        return;
+    }
+    log.error(LOG_AREA, 'An error occurred patching', { response, url: this.url });
+    onReadyToPerformNextAction.call(this);
```
lukeapage: and like in success do not call this..

src/openapi/streaming/subscription.js
```diff
@@ -351,6 +404,7 @@ Subscription.prototype.reset = function() {
             break;
 
         case STATE_SUBSCRIBED:
+            this.onUnsubscribe();
```
lukeapage: we should also do this if in the state subscribe_requested right?

src/openapi/streaming/subscription.js
```diff
@@ -392,15 +446,22 @@ Subscription.prototype.onSubscribe = function(modify) {
...
+Subscription.prototype.onModify = function(args, options) {
 
     if (this.isDisposed) {
         throw new Error('Modifying a disposed subscription - you will not get data');
     }
 
-    this.onUnsubscribe();
-    this.subscriptionData.Arguments = args;
-    this.onSubscribe(true);
+    if (options && options.method === 'PATCH') {
```
lukeapage: i'd prefer options.isPatch since there will not be multiple methods and it doesn't require a patch constant

src/openapi/streaming/subscription-queue.js
```diff
@@ -29,38 +29,67 @@ function SubscriptionQueue(maxSize) {
...
-            task === ACTION_UNSUBSCRIBE && nextTask === ACTION_SUBSCRIBE ||
-            task === ACTION_SUBSCRIBE && nextTask === ACTION_MODIFY_SUBSCRIBE ||
-            task === ACTION_MODIFY_SUBSCRIBE && nextTask === ACTION_UNSUBSCRIBE ||
-            task === nextTask
-        ) {
-            this.items.splice(i, 1);
-        }
+    if (action === ACTION_UNSUBSCRIBE ||
+        action === ACTION_SUBSCRIBE) {
+        this.reset();
```
lukeapage: if there is a modify in the queue (say we are waiting for subscribe to happen) and we get a subscribe (or an unsubscribe then a subscribe), won't this forget the fact we need to do the modify?

src/openapi/streaming/subscription-queue.js
```diff
@@ -29,38 +29,67 @@ function SubscriptionQueue(maxSize) {
...
 
+    this.items.push(queuedAction);
+
     if (this.items.length > this.maxSize) {
         // Removes elements from the beginning of a queue, to match maximum allowed size.
         this.items.splice(0, this.items.length - this.maxSize);
     }
 };
 
+SubscriptionQueue.prototype.canEnqueueAction = function canEnqueueAction(nextAction) {
```
lukeapage: if this is only used in this file, you can make the function private?

lukeapage: also no need to name it..

src/openapi/streaming/subscription-queue.js
```diff
@@ -29,38 +29,67 @@ function SubscriptionQueue(maxSize) {
...
+
+    // Removing subscribe in such case to keep requested subscription modification.
+    if (this.containsAction(ACTION_MODIFY_SUBSCRIBE) && nextAction === ACTION_SUBSCRIBE) {
+        return false;
+    }
+    return true;
+};
+
+SubscriptionQueue.prototype.containsAction = function(action) {
```
lukeapage: again, only private?

src/openapi/streaming/subscription.js
```diff
@@ -95,15 +112,17 @@ function onReadyToPerformNextAction() {
...
     performAction.call(this, this.queue.dequeue());
 }
 
 /**
  * Performs an action to a subscription based on the current state.
  * @param action
+ * @param arguments
  */
-function performAction(action) {
+function performAction(queuedActiond) {
```
lukeapage: typo in the param name?

src/openapi/streaming/subscription.js
```diff
@@ -120,6 +139,17 @@ function performAction(action) {
             }
             break;
 
+        case ACTION_MODIFY_PATCH:
+            switch (this.currentState) {
```
lukeapage: we could be subscribing, wouldn't that be ok? (waiting for initial response)

src/openapi/streaming/subscription.js
```diff
@@ -351,6 +401,8 @@ Subscription.prototype.reset = function() {
             break;
 
         case STATE_SUBSCRIBED:
+        case STATE_PATCH_REQUESTED:
+            this.onUnsubscribe();
```
lukeapage: if a patch is in process and a reset happens, then the handler for the patch will set the state to subscribed when it comes back, even if we are in a subscribe requested.. it just feels a bit fishy here..

src/openapi/streaming/subscription-queue.js
AlexeyKropotin: We do this check if (!this.canEnqueueAction(action)) before line 41

src/openapi/streaming/subscription.js
AlexeyKropotin: Then we would be in transition state, performAction is only called when we are not in a transition state

src/openapi/streaming/subscription-queue.js
```diff
@@ -29,38 +29,67 @@ function SubscriptionQueue(maxSize) {
  * - duplicates.
  * - ACTION_SUBSCRIBE action before ACTION_UNSUBSCRIBE.
  *
- * @param {Number} action - action to add to the queue.
+ * @param {Object} queuedAction - action with arguments to add to the queue.
  */
-SubscriptionQueue.prototype.enqueue = function(action) {
-    this.items.push(action);
+SubscriptionQueue.prototype.enqueue = function(queuedAction) {
```
michax: Nice performance improvement!

src/openapi/streaming/subscription.js
```diff
@@ -190,7 +220,7 @@ function onSubscribeSuccess(referenceId, result) {
     }
 
     // do not fire events if we are waiting to unsubscribe
-    if (this.queue.peek() !== ACTION_UNSUBSCRIBE) {
+    if (get(this.queue.peek(), 'action') !== ACTION_UNSUBSCRIBE) {
```
michax: What do you think about adding peekAction() method to subscription-queue?

src/openapi/streaming/subscription.js
AlexeyKropotin: We need it for reset scenario. Added unit tests for that.

src/openapi/streaming/subscription.js
AlexeyKropotin: ReferenceId change check returned

src/openapi/streaming/subscription.js
AlexeyKropotin: Done

src/openapi/streaming/subscription.js
AlexeyKropotin: Done

src/openapi/streaming/subscription-queue.js
AlexeyKropotin: Done

src/openapi/streaming/subscription-queue.js
AlexeyKropotin: Done

src/openapi/streaming/subscription.js
```diff
@@ -263,6 +292,36 @@ function onUnsubscribeError(referenceId, response) {
 }
 
 /**
+ * Called after modify patch is successful
+ * @param referenceId
+ * @param response
+ */
+function onModifyPatchSuccess(referenceId, response) {
+    if (referenceId !== this.referenceId) {
+        log.error(LOG_AREA, 'Received a response for modify patch a subscription that has afterwards been reset - ignoring');
```
lukeapage: this should not be an error.. warning?
I guess it could only happen if we have a reset?

src/openapi/streaming/subscription.js
```diff
@@ -351,6 +410,8 @@ Subscription.prototype.reset = function() {
             break;
 
         case STATE_SUBSCRIBED:
+        case STATE_PATCH_REQUESTED:
```
lukeapage: doesn't onStreamingData need to also go ahead if in state patch requested?

lukeapage: same for timeTillOrphaned ?

test/unit/openapi/streaming/subscription-queue-spec.js
```diff
@@ -11,106 +11,191 @@ describe("openapi SubscriptionQueue", () => {
...
+
+            expect(queue.dequeue().action).toBe(SubscriptionActions.ACTION_UNSUBSCRIBE);
+            expect(queue.dequeue().action).toBe(SubscriptionActions.ACTION_MODIFY_SUBSCRIBE);
+            expect(queue.isEmpty()).toBe(true);
+        })
+
+        it('should insert modify subscribe and remove any non unsubscribe actions', () => {
+            queue.enqueue({ action: SubscriptionActions.ACTION_MODIFY_PATCH });
+            queue.enqueue({ action: SubscriptionActions.ACTION_MODIFY_PATCH });
+            queue.enqueue({ action: SubscriptionActions.ACTION_MODIFY_SUBSCRIBE });
```
lukeapage: this would never happen right? you would have an unsubscribe before this modify...

lukeapage: either delete the test or comment on it

src/openapi/streaming/subscription-queue.js
```diff
@@ -21,60 +21,113 @@ import {
...
-            this.items.splice(i, 1);
+    const { action } = queuedItem;
+    const prevAction = this.getLastItem().action;
+
+    // ..UM => UM
+    if (action === ACTION_MODIFY_SUBSCRIBE) {
+        if (prevAction === ACTION_UNSUBSCRIBE) {
+            this.items = [{ action: ACTION_UNSUBSCRIBE }, queuedItem];
+        } else {
+            this.items = [queuedItem];
```
AlexeyKropotin: That is right

src/openapi/streaming/subscription.js
AlexeyKropotin: done

src/openapi/streaming/subscription-queue.js
```diff
@@ -21,60 +21,111 @@ import {
  */
 function SubscriptionQueue(maxSize) {
     this.items = [];
-    this.maxSize = maxSize || 2;
+
+    this.getLastItem = function() {
```
lukeapage: why isn't this on the prototype with jsdoc?

AlexeyKropotin: done

PR #29, opened 2017-04-12
add options of status codes and custom timeouts to TransportRetry

ilyalopatkin:
Extends the TransportRetry to also apply to successful requests that returned a given status code (e.g. 504 Gateway timeout). Adds custom timeouts configuration to individual method option.

### DIFF
--- a/src/openapi/transport/retry.js
+++ b/src/openapi/transport/retry.js
@@ -10,7 +10,7 @@
 function transportMethod(method) {
     return function() {
         // checking if http method call should be handled by RetryTransport
-        if (this.methods[method] && this.methods[method].retryLimit > 0) {
+        if (this.methods[method] && (this.methods[method].retryLimit > 0 || this.methods[method].retryTimeouts)) {
 
             return new Promise((resolve, reject) => {
 
@@ -20,6 +20,7 @@ function transportMethod(method) {
                     resolve,
                     reject,
                     retryCount: 0,
+                    retryTimer: null,
                 };
 
                 this.sendTransportCall(transportCall);
@@ -44,7 +45,13 @@ function transportMethod(method) {
  * @param {object.<string,object>} [options.methods] - Http methods that should retry. For each method provide an object with `retryLimit` parameter.
  * @example
  * // Constructor with parameters
- * var transportRetry = new TransportRetry(transport, {retryTimeout:10000, methods:{'delete':{retryLimit:3}}});
+ * var transportRetry = new TransportRetry(transport, {
+ *      retryTimeout:10000,
+ *      methods:{
+ *          'delete':{ retryLimit:3 },
+ *          'post':{ retryTimeouts: [1000, 1000, 2000, 3000, 5000], statuses: [504] },
+ *      }
+ * });
  */
 function TransportRetry(transport, options) {
     if (!transport) {
@@ -116,8 +123,10 @@ TransportRetry.prototype.sendTransportCall = function(transportCall) {
         .apply(this.transport, transportCall.args)
         .then(transportCall.resolve,
             (response) => {
-                if (!(response && response.status) &&
-                    (transportCall.retryCount < this.methods[transportCall.method].retryLimit)) {
+                const callOptions = this.methods[transportCall.method];
+                if ((!(response && response.status) || callOptions.statuses && callOptions.statuses.indexOf(response.status) >= 0) &&
+                    (callOptions.retryLimit > 0 && transportCall.retryCount < callOptions.retryLimit ||
+                        callOptions.retryTimeouts && transportCall.retryCount < callOptions.retryTimeouts.length)) {
                     this.addFailedCall(transportCall);
                 } else {
                     transportCall.reject.apply(null, arguments);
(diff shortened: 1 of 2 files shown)


src/openapi/transport/retry.js
```diff
@@ -116,8 +123,10 @@ TransportRetry.prototype.sendTransportCall = function(transportCall) {
         .apply(this.transport, transportCall.args)
         .then(transportCall.resolve,
             (response) => {
-                if (!(response && response.status) &&
-                    (transportCall.retryCount < this.methods[transportCall.method].retryLimit)) {
+                const callOptions = this.methods[transportCall.method];
+                if ((!(response && response.status) || callOptions.statuses && callOptions.statuses.indexOf(response.status) >= 0) &&
```
lukeapage: perhaps it would be worth splitting out this if and assigning to variables first.. because there is now a lot of conditions, so it is difficult to follow.

src/openapi/transport/retry.js
```diff
@@ -130,29 +139,63 @@ TransportRetry.prototype.sendTransportCall = function(transportCall) {
...
-    if (!this.retryTimer) {
+    const callOptions = this.methods[transportCall.method];
+    if (callOptions.retryTimeouts && callOptions.retryTimeouts.length > transportCall.retryCount) {
+        // schedule an individual retry timeout
+        transportCall.retryTimer = setTimeout(() => {
+            this.retryIndividualFailedCall(transportCall);
+        }, callOptions.retryTimeouts[transportCall.retryCount]);
+    } else if (!this.retryTimer) {
+        // schedule a retry timeout for all failed calls
         this.retryTimer = setTimeout(() => {
```
lukeapage: don't we need to not add it to failedCalls? otherwise it will retry both

lukeapage: ah I see.. you skip it.. but would it be simpler to not add it? do we use that array for anything else?

src/openapi/transport/retry.js
```diff
@@ -130,29 +139,63 @@ TransportRetry.prototype.sendTransportCall = function(transportCall) {
...
+    this.sendTransportCall(transportCall);
+};
+
+/**
+ * Disposes the underlying transport, the failed calls queue and clears retry timers.
  */
 TransportRetry.prototype.dispose = function() {
+    this.failedCalls.forEach((failedCall) => {
+        if (failedCall.retryTimer) {
+            failedCall.retryTimer.clear();
```
lukeapage: isn't retryTimer the result of setTimeout?

src/openapi/transport/retry.js
ilyalopatkin: No, it's just to hold all failed/pending calls. My thought was to keep all failed calls in the same array (its name kinda suggested that) and figure out when they fire separately. We better keep the individually timed ones somewhere so we clean up properly, but that could be in a separate array if you think it's more straightforward?

src/openapi/transport/retry.js
ilyalopatkin: ah it needs to be clearTimeout(), there's a little bug below this then where it does this.retryTimer.clear()

src/openapi/transport/retry.js
lukeapage: I don’t think we should add it to the array of failed calls
because the only use it has is we have to filter it out when retryFailedCalls is called

src/openapi/transport/retry.js
```diff
@@ -44,7 +45,13 @@ function transportMethod(method) {
  * @param {object.<string,object>} [options.methods] - Http methods that should retry. For each method provide an object with `retryLimit` parameter.
  * @example
  * // Constructor with parameters
- * var transportRetry = new TransportRetry(transport, {retryTimeout:10000, methods:{'delete':{retryLimit:3}}});
+ * var transportRetry = new TransportRetry(transport, {
+ *      retryTimeout:10000,
```
michax: Post merge question :)
How this retryTimeout relates to ones declared in methods?
Is this default if retryTimeouts are not provied for given methods?

src/openapi/transport/retry.js
```diff
@@ -150,12 +171,31 @@ TransportRetry.prototype.retryFailedCalls = function() {
...
+    this.individualFailedCalls.forEach((failedCall) => {
+        if (failedCall.retryTimer) {
+            failedCall.retryTimer.clearTimeout();
+            failedCall.retryTimer = null;
+        }
+    });
     this.failedCalls.length = 0;
     if (this.retryTimer) {
-        this.retryTimer.clear();
+        this.retryTimer.clearTimeout();
```
michax: Hmm I don't understand this part. Afaik, this.retryTimer holds result of set setTimeout, which normaly returns number not instance with methods like clear and clearTimeout. 

I think I'm missing something :) Do we assign something else the setTimeout result to this property ?

src/openapi/transport/retry.js
ilyalopatkin: yes that's the default for all methods

src/openapi/transport/retry.js
ilyalopatkin: No you're right... it should be clearTimeout(this.retryTimer), I don't know what made me think it's an object, and apparently this.retryTimer.clear() wasn't doing much. Basically dispose was and still is broken, I guess it's not critical but worth fixing for next version

PR #31, opened 2017-04-26
clear timeouts and reject outstanding promises in TransportRetry dispose

ilyalopatkin:


src/openapi/transport/retry.js
```diff
@@ -128,7 +128,7 @@ TransportRetry.prototype.sendTransportCall = function(transportCall) {
                 const isRetryRequest = !(response && response.status) || callOptions.statuses && callOptions.statuses.indexOf(response.status) >= 0;
                 const isWithinRetryLimitOption = callOptions.retryLimit > 0 && transportCall.retryCount < callOptions.retryLimit;
                 const isWithinRetryTimeoutsOption = callOptions.retryTimeouts && transportCall.retryCount < callOptions.retryTimeouts.length;
-                if (isRetryRequest && (isWithinRetryLimitOption || isWithinRetryTimeoutsOption)) {
+                if (isRetryRequest && (isWithinRetryLimitOption || isWithinRetryTimeoutsOption) && !this.isDisposed) {
```
lukeapage: Why not return above this?

ilyalopatkin: this is to reject the calls that were in progress when transport disposed

lukeapage: I see.. so it fail the call immediately.

PR #35, opened 2017-05-10
Modify patch arguments

AlexeyKropotin:


test/unit/openapi/streaming/subscription-spec.js
```diff
@@ -807,20 +814,19 @@ describe("openapi StreamingSubscription", () => {
             sendInitialResponse({InactivityTimeout: 100, Snapshot: { resetResponse: true }});
 
             tick(() => {
-                const firstArgs = { newArgs: 'firstArgs' };
-                const secondArgs = { newArgs: 'secondArgs' };
-                subscription.onModify(firstArgs, { isPatch: true });
-                subscription.onModify(secondArgs, { isPatch: true });
+                const newArgs = { newArgs: 'test' };
```
lukeapage: this is changing the test - shouldn't it pass the same thing as before - and we test that the last args is used?

test/unit/openapi/streaming/subscription-spec.js
```diff
@@ -807,20 +814,19 @@ describe("openapi StreamingSubscription", () => {
...
-                expect(transport.patch.calls.all()[0].args[3].body).toEqual(firstArgs);
+                expect(transport.patch.calls.all()[0].args[3].body).toEqual({ newArgs: 'firstArgs' });
 
                 transport.patchResolve({ status: "200", response: "" });
 
                 tick(() => {
                     expect(transport.patch.calls.count()).toEqual(2);
                     // second patch arguments sent
-                    expect(transport.patch.calls.all()[1].args[3].body).toEqual(secondArgs);
+                     expect(transport.patch.calls.all()[1].args[3].body).toEqual({ newArgs: 'secondArgs' });
```
lukeapage: extra additional space before expect?

src/openapi/streaming/subscription.js
```diff
@@ -456,23 +456,26 @@ Subscription.prototype.onSubscribe = function(modify) {
...
+            throw new Error('Modify options patchArgsDelta is not defined');
+        }
+
+        this.subscriptionData.Arguments = newArgs;
+        tryPerformAction.call(this, ACTION_MODIFY_PATCH, options.patchArgsDelta);
     } else {
         // resubscribe with new arguments
         this.onUnsubscribe();
-        this.subscriptionData.Arguments = args;
+        this.subscriptionData.Arguments = newArgs;
```
lukeapage: you can do this at the top, now it is the same for both types

PR #37, opened 2017-05-23
Send response instead of request arguments in transportRetry reject case

sankalplakhina:


src/openapi/transport/retry.js
```diff
@@ -131,7 +131,7 @@ TransportRetry.prototype.sendTransportCall = function(transportCall) {
                 if (isRetryRequest && (isWithinRetryLimitOption || isWithinRetryTimeoutsOption) && !this.isDisposed) {
                     this.addFailedCall(transportCall);
                 } else {
-                    transportCall.reject.apply(null, arguments);
+                    transportCall.reject.call(null, response);
```
lukeapage: this can now be 
```
transportCall.reject(response);
```

lukeapage: Also add unit tests that would fail without this change..

sankalplakhina: Updates the PR with changes.

PR #39, opened 2017-06-19
open api no longer require the active part of the url

lukeapage:


test/unit/openapi/streaming/subscription-spec.js
```diff
@@ -37,14 +37,14 @@ describe("openapi StreamingSubscription", () => {
...
-            expect(transport.post.calls.argsFor(0)).toEqual(['serviceGroup', 'test/resource/active', null, jasmine.objectContaining({ body: jasmine.objectContaining({ RefreshRate: 120 }) })]);
+            expect(transport.post.calls.argsFor(0)).toEqual(['serviceGroup', 'test/resource', null, jasmine.objectContaining({ body: jasmine.objectContaining({ RefreshRate: 120 }) })]);
         });
         it("has a minimum refresh rate", () => {
             var subscription = new Subscription('123', transport, 'serviceGroup', 'test/resource', { RefreshRate: 1 });
             subscription.onSubscribe();
 
             expect(transport.post.calls.count()).toEqual(1);
-            expect(transport.post.calls.argsFor(0)).toEqual(['serviceGroup', 'test/resource/active', null, jasmine.objectContaining({ body: jasmine.objectContaining({ RefreshRate: 100 }) })]);
+            expect(transport.post.calls.argsFor(0)).toEqual(['serviceGroup', 'test/resource', null, jasmine.objectContaining({ body: jasmine.objectContaining({ RefreshRate: 100 }) })]);
```
michax: Nitpick. We are testing transport.post for subscribe. However we don't validate it for patches. Maybe it would be worth adding same unit test, that will assert built url for patch. However, it could be task for another PR :)

lukeapage: its a good point..

lukeapage: Done

PR #45, opened 2017-07-14
Protobuf serialization

michax:
This change introduces support for protobuf serialization.
It's driven by subscription Format parameter.

**Stats**

Current library size increase. Around 83 KB minified and 21 KB gziped.
In average, we safe around 419 KB per 1000 updates (based on balances subscription).
Minimal CPU hit.

**Changes**

- Using protobufjs library. Responsibilites: decoding, runtime parsing of incoming schemas, reflection api for parsed schemas. Support for additive parsing of schemas.
- Custom metadata processor, required for correct handling of delta updates
- Per endpoint schema storage. Required due to lack of namespaces in received schemas and duplicated message names with different structures.
- Wrapper for google.Timestamp (converting Timestamp to Date)
- Full unit test coverage for, with example schemas and encoded protobuf messages.
- Each update is still enveloped in JSON message, due to signalR limitation (no possibility to easily set custom websocket data type ie. ArrayBuffer).
- Each message content is passed as base64 string. 
- Minimal changes were done to subscription implementation. Just hooking up serialization per Format type.


src/openapi/streaming/serializer/schemas/proto-base.js
```diff
@@ -0,0 +1,70 @@
+// Stores core schemas used by all schemas received from openapi.
+
+export const protoBclBase = `package bcl;
```
michax: Probably not needed if we decide to go with pure google types without C# ones like Decimal.

src/openapi/streaming/serializer/schemas/proto-base.js
```diff
@@ -0,0 +1,70 @@
...
+  optional fixed64 hi = 2; // the second 8 bytes of the guid
+}
+
+message Decimal {
+  optional uint64 lo = 1; // the first 64 bits of the underlying value
+  optional uint32 hi = 2; // the last 32 bis of the underlying value
+  optional uint32 signScale = 3; // the number of decimal digits (bits 1-16), and the sign (bit 0)
+}`;
+
+export const protoGoogleBase = `
```
michax: We should collect all required core types that we need. Depending on the amount, it could impact library size.

Alternative would be to always return core schemas in first subscription request, and provide them as KnownSchemas. 
However, this could cause issues if we request multiple subscriptions, and all of them will get initially schemas with core definitions.

Last option would be initial negotiation for core schemas.

src/openapi/streaming/serializer/serializer-protobuf.js
```diff
@@ -0,0 +1,125 @@
+import * as baseProtos from './schemas/proto-base';
+import protobuf from 'protobufjs/dist/protobuf';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import base64 from '@protobufjs/base64';
```
michax: namespaced export from protobufjs. Not sure why it's detected as an error by eslint. By using built in base64 from protobufjs we save some space by not using separate base64 implementation.

src/openapi/streaming/serializer/serializer-protobuf.js
```diff
@@ -0,0 +1,125 @@
...
+import log from '../../../log';
+
+const LOG_AREA = 'SerializerProtobuf';
+
+/**
+ * Protobuf Serialization
+ * @constructor
+ */
+function SerializerProtobuf() {
```
michax: Should I create interface (base serializer) for serializers? It would state clear contract for all serializers.

src/openapi/streaming/subscription.js
```diff
@@ -1,4 +1,4 @@
-/* eslint max-lines: ["error", 600] */
+/* eslint max-lines: ["error", 650] */
```
michax: Minimal amount of new code was added to subscription. Actually no new logic, only invocation of encapsulated serializers and comments.

New line count is 623. So if I would need to keep max 600, refactoring of subscription.js outside my changes is required.

test/unit/mocks/proto-price.js
```diff
@@ -0,0 +1,271 @@
+// Size summary
+// JSON with numeric fields: 1086 bytes
+// JSON: 1556
+// Protobuf base64: 800 bytes
+// Probobuf ByteArray: 600 bytes
+
+export const schema = `syntax = "proto3";
```
michax: Price Response schema provided by openapi. Quite heavy one. Acts as good testing dummy.

test/unit/openapi/streaming/subscription-spec.js
```diff
@@ -752,6 +753,181 @@ describe('openapi StreamingSubscription', () => {
         });
     });
 
+    describe('protofub serialization', () => {
```
michax: protofub -> protobuf

src/openapi/streaming/serializer/serializer-protobuf.js
michax: Not needed anymore. In latest version, util is available in protobuf.

src/openapi/streaming/serializer/schemas/proto-base.js
michax: We agreed upon requirements for additional types. For now, we only use google.Timestamp. Which has built in schema inside protobufjs. So no need for this file anymore. 

As for C# oriented types, we decided to not use them. So instead of Decimal, we just use native Double which is mapped to Number.

src/openapi/streaming/serializer/schemas/proto-base.js
michax: Yeap, removed.

package.json
```diff
@@ -35,12 +36,15 @@
     "grunt-contrib-copy": "1.0.0",
     "grunt-contrib-jasmine": "1.0.3",
     "grunt-contrib-watch": "1.0.0",
-    "grunt-rollup": "1.0.1",
+    "grunt-rollup": "6.0.0",
```
michax: I updated build related packages in hopes we will gain some size improvements. But in the end it didn't help. 
Nevertheless it works fine with update.

src/openapi/streaming/serializer/meta/meta-protobuf.js
```diff
@@ -0,0 +1,103 @@
...
+function MetaProtobuf() {}
+
+/**
+ * Process data using message metadata. Iterate through each field and process metadata, if found.
+ *
+ * @param {Object} message - Protobuf Message Type object.
+ * @param {Object} data - JSON object. Object get's mutated.
+ * @return {*}
+ */
+MetaProtobuf.prototype.process = function(message, data) {
```
michax: More on why it's needed here:
https://wiki/display/OpenAPI/Delta+compression+implementation+of+ProtoBuffers

Sadly at this point, protobufjs doesn't offer any hooks for post processing data. So we need to do it manually here.
I've tested performance and it has minimal hit (0.01 ms level).

src/openapi/streaming/serializer/serializer-protobuf.js
```diff
@@ -0,0 +1,153 @@
...
+    const rootType = this.getSchemaType(schemaName, rootTypeName);
+
+    if (!rootType) {
+        log.error('Parsing failed. Root type not found. Name: ', rootTypeName);
+        return null;
+    }
+
+    const byteArray = new Uint8Array(protobuf.util.base64.length(data));
+
+    try {
```
michax: Try and catch slows down things a little bit. 
What do you think, should we keep it, or allow protobufjs errors to go to the top? 

It's performance critical method, I think we should get rid of.

src/openapi/streaming/subscription.js
```diff
@@ -372,23 +374,34 @@ function setState(state) {
...
+     * The action queue.
      * @type {SubscriptionQueue}
      */
     this.queue = new SubscriptionQueue();
 
+    /**
+     * The subscription format. Used to derive serialization method for incoming subscription messages.
+     * @type {String}
+     */
+    this.subscriptionFormat = subscriptionArgs.Format;
```
michax: Drop subscription prefix from subscriptionFormat.

src/openapi/streaming/serializer/serializer-json.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * JSON Serialization
+ * @constructor
+ */
+function SerializerJson(name) {
```
michax: Consider creating interface for both Serialization methods we now support. However, JSON at this point is just push through, as singlalR already parses to JSON.

src/openapi/streaming/serializer/serializer-protobuf.js
```diff
@@ -0,0 +1,153 @@
...
+ * @return {*} - Result of parsing, if successful. Returns null if parsing fails or there is no data.
+ */
+SerializerProtobuf.prototype.parse = function(data, schemaName) {
+    const schemas = this.getSchema(schemaName);
+
+    if (!schemas || !data) {
+        return null;
+    }
+
+    const rootTypeName = schemas.root.getOption('saxobank_root');
```
michax: Move ''saxobank_root' to const. ie. ROOT_OPTION_NAME.

src/openapi/streaming/serializer/serializer-protobuf.js
```diff
@@ -0,0 +1,155 @@
...
+
+    if (!rootType) {
+        log.error('Parsing failed. Root type not found. Name: ', rootTypeName);
+        return null;
+    }
+
+    const byteArray = new Uint8Array(protobuf.util.base64.length(data));
+
+    try {
+        protobuf.util.base64.decode(data, byteArray, 0);
```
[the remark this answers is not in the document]
michax: It's offset, will do.

src/openapi/streaming/serializer-facade.js
```diff
@@ -0,0 +1,53 @@
...
+    if (SerializerFacade.serializersMap[id]) {
+        return SerializerFacade.serializersMap[id];
+    }
+    const Serializer = SerializerFacade.serializerCreators[format] || SerializerFacade.defaultSerializer;
+    SerializerFacade.serializersMap[id] = new Serializer(id);
+
+    return SerializerFacade.serializersMap[id];
+};
+
+window.$serializerFacade = SerializerFacade;
```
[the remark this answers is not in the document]
michax: Good catch, leftover from debugging session.

src/openapi/streaming/subscription.js
```diff
@@ -442,6 +455,21 @@ Subscription.prototype.removeStateChangedCallback = function(callback) {
     }
 };
 
+Subscription.prototype.processUpdate = function(message, type) {
+    message.Data = this.serializer.parse(message.Data, this.SchemaName);
```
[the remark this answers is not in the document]
michax: I heard legends about it but for protobuf with current approach we cannot support it, and openapi, i hope took that into account. Will confirm with them.

src/openapi/streaming/serializer/serializer-protobuf.js
michax: I'll check using some unit tests. And derive some tangible cost of try/catch.

src/openapi/streaming/subscription.js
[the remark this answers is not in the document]
michax: Ok, looking at the implementation, this is already handled.
We store delta updates inside updatesBeforeSubscribed, if they arrived before snapshot. So we should be good here.

src/openapi/streaming/subscription.js
```diff
@@ -1,4 +1,4 @@
-/* eslint max-lines: ["error", 700] */
+/* eslint max-lines: ["error", 750] */
```
lukeapage: maybe we hsould just disable this ;)

src/openapi/streaming/subscription.js
```diff
@@ -372,23 +374,34 @@ function setState(state) {
...
+     * The action queue.
      * @type {SubscriptionQueue}
      */
     this.queue = new SubscriptionQueue();
 
+    /**
+     * The subscription format. Used to derive serialization method for incoming subscription messages.
+     * @type {String}
+     */
+    this.format = subscriptionArgs.Format;
```
lukeapage: how do we stop people forgetting a Format argument? maybe we can wrap streaming and always provide it?

src/openapi/streaming/subscription.js
```diff
@@ -442,6 +455,21 @@ Subscription.prototype.removeStateChangedCallback = function(callback) {
...
+    this.onUpdate(message, type);
+};
+
+Subscription.prototype.processSnapshot = function(response) {
+    if (response.Schema && response.SchemaName) {
+        this.SchemaName = response.SchemaName;
+        this.serializer.addSchema(response.Schema, response.SchemaName);
+    }
+
+    // Snapshot protobuf serialization is currently 'nice to have' on openapi side. So, for now, always using JSON.
```
lukeapage: more professional comment maybe?
// Snapshot protobuf serialization is currently not supported
?

src/openapi/streaming/serializer/serializer-protobuf.js
```diff
@@ -0,0 +1,153 @@
+import MetaProcessor from './meta/meta-protobuf';
+import wrappers from './wrappers/protobuf-wrappers';
+import protobuf from 'protobufjs/dist/protobuf.min';
```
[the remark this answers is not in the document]
michax: It produces smaller openapi-client lib size, but in the end, when minified it should produce same size as with this solution. So it's less scary in non-minified version. Other then that no advantages. 

As for using webpack instead of rollup. Not sure if it should be done as part of this PR. But in the end, as Webpack doesn't support tree shaking, I doubt it would preduce better results.

grunt/rollup-conf.js
```diff
@@ -32,6 +34,8 @@ module.exports = {
             banner: outputParts[0],
             footer: outputParts[1],
             plugins: [
```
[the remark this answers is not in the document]
michax: I thought about having 2 builds and using some pre-processing etc. But it seemed fugly.


What about exposing method for setting protobuf implementation from openapi-clientlib user space? 
To support protbuf, we would import protobuf implementation on **go** side and pass it to openapi-clientlib. 


If protobuf will not be passed, we will force standard, JSON serialization method.


Are there any other openapi-client lib users that want to benefit from protobuf, or go is main and only userspace atm? 

We already require external dependencies ie. SignalR and Jquery. So precedence is there. However, in this case it would be configurable optional external dependency. 

I think this allows for more flexibility. In future we could detect passed protobuf implementation and support few of them. 

Example usage

```
import protobuf from 'protobufjs';
...
streamingOptions = {
    transportTypes: ['longPolling'],
    serializationFormats: [
         'application/x-protobuf': protobuf,
    ],
}
...
openApi.streaming = new openapi.Streaming(retryTransport, openApiConfig.streamingUrl, transportAuth.auth, streamingOptions)
```

src/openapi/streaming/serializer/serializer-protobuf.js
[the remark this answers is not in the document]
michax: As discussed, let try to implement passing of protobuf from openapi-clientlib userspace.

src/openapi/streaming/serializer-facade.js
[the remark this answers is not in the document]
michax: As a punishment I could bring eslint plugin for window. It shouldn't be used generally and if there will be a need it will be an exception.

grunt/rollup-conf.js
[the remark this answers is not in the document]
michax: As discussed, added API for defining external serializers and underlying engines like protobufjs.

In addition to defining serializer per format, user is able to set underlying engine.

src/openapi/streaming/subscription.js
michax: To be honest, I would leave it be. It's good reminder to try a encapsulate some close coupled functionalities. 

However, I would happily change default to 900 from 300, globally.

In the end, lines per functions would be better indicator of 'clean code'.

src/openapi/streaming/serializer/serializer-interface.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Serialization Interface
+ * @constructor
+ */
+function SerializerInterface(name, engine = null) {
```
jawalker88: unusual to have implementation in an "interface"? same with getFormatName below? I think both implementations do their own thing anyway and don't use these?

michax: Yeah, Interface wasn't best name. I just hate suffixing/prefixing it with 'Base', and Serializer alone seems vague. 

Will change it to Base however, is if we would like to follow interface requirements, We should implement all methods.

jawalker88: Base makes most sense to me personally if it contains implementation.
Contract sounds like interface to me, I wouldn't expect any implementation. I'm not particularly tied to either though, as long as it's not interface :)

michax:
Removed commonjs and node resolver plugins from dist build. However I left it for test build as we need it for testing protobuf serialization.

grunt/rollup-conf.js
```diff
@@ -8,6 +8,8 @@ var packageConfig = require("../package.json");
 var sharedConfig = require("./_shared-config");
 var fs = require("fs");
 var babel = require("rollup-plugin-babel");
+var resolve = require("rollup-plugin-node-resolve");
```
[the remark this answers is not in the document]
michax: Only used by tests, for importing protobufjs. It's not used when building dist, and userspace is required to import protobufjs in their scope.

package.json
```diff
@@ -35,12 +36,15 @@
...
     "grunt-template-jasmine-istanbul": "0.4.0",
     "jquery": "2.1.4",
+    "protobufjs": "6.8.0",
```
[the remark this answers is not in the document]
michax: Yes, userspace will be required to import it in their scope and provide as openapi factory configuration. It still present as dev dependency because it's used in unit tests.

PR #46, opened 2017-07-22
Add requestCounter to identify unique request for every open api rest…

grajiv:
Added RequestCounter for every rest request to identify unique request.

### DIFF
--- a/src/openapi/transport/core.js
+++ b/src/openapi/transport/core.js
@@ -9,7 +9,7 @@ import fetch from '../../utils/fetch';
 // -- Local variables section --
 
 const DEFAULT_CACHE = true;
-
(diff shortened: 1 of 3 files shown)


src/openapi/transport/core.js
```diff
@@ -75,6 +80,8 @@ function Transport(baseUrl, options) {
     this.baseUrl = baseUrl;
     this.language = options && options.language;
     this.defaultCache = options && typeof options.defaultCache === 'boolean' ? options.defaultCache : DEFAULT_CACHE;
+
```
lukeapage: I think this should be outside the class so if you create 2 transports they share the same counter..

grajiv: Done

PR #49, opened 2017-08-09
allow fetch to handle binary responses when content-type is a pdf or an excel sheet

staunsholm:
This will allow fetch to handle binary responses when content-type is a pdf or an excel sheet.

### DIFF
--- a/src/utils/fetch.js
+++ b/src/utils/fetch.js
@@ -9,6 +9,12 @@ import log from '../log';
 
 const LOG_AREA = 'Fetch';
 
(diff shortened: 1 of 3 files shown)


src/utils/fetch.js
```diff
@@ -106,6 +112,16 @@ function convertFetchResponse(url, body, result, isRejected) {
                 url,
             };
         });
+    } else if (contentType && binaryContentTypes[contentType] > -1) {
```
lukeapage: merge into the if above?
Add a test..

PR #50, opened 2017-08-14
Add support for decimal values to NumberFormatting.shortFormat

### LINKED ISSUE #48: Floating point issue
There's a floating point issue in the format methods:
```
numberFormat.shortFormat(10.1, 2)
> "10"
numberFormat.shortFormat(10.11, 2)
> "0.0k"
```
Will investigate a fix and make a PR.

Pephers:
NumberFormatting.shortFormat didn't support decimals properly. Specifically, some floating points gave issues such as 1000000.11 => 0m. This fixes the issue by calculating the boundary from digit size.

Fixes #48

### DIFF
--- a/src/number-formatting/short-format.js
+++ b/src/number-formatting/short-format.js
@@ -18,23 +18,23 @@ import formatNumber from './format';
  * @returns {string} Returns 0 when dates are equal. -1 when date1 less than date2. 1 when date1 greater than date2.
  */
 function shortFormat(num, options) {
-
-    let numberSize = String(num).length; // Unfortunately Logs are too inaccurate - Math.round(Math.log(num) / Math.LN10)
+    const [digits] = String(num).split('.');
+    let digitSize = digits.length;
     let boundary;
 
-    if (numberSize >= 5) { // bigger than 10,000
-        boundary = Math.pow(10, numberSize) - (Math.pow(10, numberSize - 3) / 2); // e.g. 100,000 -> 9,9950 - closer to 100k than 99.9k
+    if (digitSize >= 5) { // bigger than 10,000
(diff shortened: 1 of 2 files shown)


src/number-formatting/short-format.js
```diff
@@ -18,26 +18,29 @@ import formatNumber from './format';
  * @returns {string} Returns 0 when dates are equal. -1 when date1 less than date2. 1 when date1 greater than date2.
  */
 function shortFormat(num, options) {
+    const [digits, decimals] = String(num).split('.');
+    const decimalSize = decimals ? decimals.length : 0;
+    const number = Number(`${digits}e${decimalSize}`);
```
lukeapage: Why not Number(digits)
What is this doing? 

105.232 => 105e3 ? == 105000 which it then compares against boundaries?

Pephers: Good point, it removes the decimal value which is definitely unintended. I'll give it a second pass.

Pephers:
I've updated the code to preserve decimal values.

Pephers:
Changed code to calculate boundary from digit size.

test/unit/number-formatting/format-spec.js
```diff
@@ -57,15 +56,25 @@ describe('NumberFormatting format', () => {
             expect(shortFormat(99949)).toEqual('99.9k');
             expect(shortFormat(99950)).toEqual('100k');
 
-            expect(shortFormat(999499)).toEqual('999k');
-            expect(shortFormat(999500)).toEqual('1.00m');
+            expect(shortFormat(999499.11)).toEqual('999k');
```
lukeapage: move this to the decimals spec and put this back?

Pephers: Fixed.

PR #54, opened 2017-10-17
Add functionality to unsubscribe by tag

jawalker88:
In progress

### DIFF
--- a/src/openapi/streaming/streaming.js
+++ b/src/openapi/streaming/streaming.js
@@ -1,3 +1,4 @@
+/* eslint max-lines: ["error", 700] */
 /**
  * @module saxo/openapi/streaming/streaming
  * @ignore
@@ -354,6 +355,91 @@ function onOrphanFound(subscription) {
     subscription.reset();
 }
 
+function handleSubscriptionReadyForUnsubscribe(subscriptions, resolve) {
+    let allSubscriptionsReady = true;
+    for (let i = 0; i < subscriptions.length && allSubscriptionsReady; i++) {
+        if (!subscriptions[i].isReadyForUnsubscribeByTag()) {
+            allSubscriptionsReady = false;
+        }
+    }
+
+    if (allSubscriptionsReady) {
+        resolve();
+    }
+}
+
+function getSubscriptionsByTag(serviceGroup, url, tag) {
+    const subscriptionsToRemove = [];
+
+    for (let i = 0; i < this.subscriptions.length; i++) {
+        const subscription = this.subscriptions[i];
+
+        if (subscription.serviceGroup === serviceGroup &&
+                subscription.url === url &&
+                subscription.subscriptionData.Tag === tag) {
+
+            subscriptionsToRemove.push(subscription);
+        }
+    }
+
+    return subscriptionsToRemove;
+}
+
+function getSubscriptionsReadyPromise(subscriptionsToRemove, shouldDisposeSubscription) {
+    let onStateChanged;
+
+    return new Promise((resolve) => {
+        onStateChanged = handleSubscriptionReadyForUnsubscribe.bind(this, subscriptionsToRemove, resolve);
+
+        for (let i = 0; i < subscriptionsToRemove.length; i++) {
+            const subscription = subscriptionsToRemove[i];
+
+            ignoreSubscriptions[subscription.referenceId] = true;
+            setTimeout(() => delete ignoreSubscriptions[subscription.referenceId], MS_TO_IGNORE_DATA_ON_UNSUBSCRIBED);
+
+            subscription.addStateChangedCallback(onStateChanged);
+            subscription.onUnsubscribeByTagPending();
+
+            if (shouldDisposeSubscription) {
+                removeSubscription.call(this, subscription);
+            }
+        }
+    })
+    .then(() => {
+        for (let i = 0; i < subscriptionsToRemove.length; i++) {
+            const subscription = subscriptionsToRemove[i];
+            subscription.removeStateChangedCallback(onStateChanged);
+        }
+    });
+}
+
+function unsubscribeSubscriptionByTag(serviceGroup, url, tag, shouldDisposeSubscription) {
+    const subscriptionsToRemove = getSubscriptionsByTag.call(this, serviceGroup, url, tag);
+    const allSubscriptionsReady = getSubscriptionsReadyPromise.call(this, subscriptionsToRemove, shouldDisposeSubscription);
+
+    allSubscriptionsReady.then(() => {
+        this.transport.delete(serviceGroup, url + '/{contextId}/?Tag={tag}', {
+            contextId: this.contextId,
+            tag,
+        })
+        .catch((response) => log.error(LOG_AREA, 'An error occurred unsubscribing by tag', { response, serviceGroup, url, tag }))
+        .then(() => {
+            for (let i = 0; i < subscriptionsToRemove.length; i++) {
+                const subscription = subscriptionsToRemove[i];
+                subscription.onUnsubscribeByTagComplete();
+            }
+        });
+    });
+}
+
+function removeSubscription(subscription) {
+    subscription.dispose();
+    const indexOfSubscription = this.subscriptions.indexOf(subscription);
+    if (indexOfSubscription >= 0) {
+        this.subscriptions.splice(indexOfSubscription, 1);
+    }
+}
+
 // -- Exported methods section --
 
 /**
@@ -451,6 +537,7 @@ Streaming.prototype.READABLE_CONNECTION_STATE_MAP = {
  * @param {number} [subscriptionArgs.RefreshRate=1000] - The data refresh rate (passed to OpenAPI).
  * @param {string} [subscriptionArgs.Format] - The format for the subscription (passed to OpenAPI).
  * @param {object} [subscriptionArgs.Arguments] - The subscription arguments (passed to OpenAPI).
+ * @param {string} [subscriptionArgs.Tag] - The tag for the subscription (passed to OpenAPI).
  * @param {function} onUpdate - A callback function that is invoked when an initial snapshot or update is received.
  *                              The first argument will be the data received and the second argument will either be
  *                              subscription.UPDATE_TYPE_DELTA or subscription.UPDATE_TYPE_SNAPSHOT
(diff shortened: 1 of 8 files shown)


src/openapi/streaming/streaming.js
```diff
@@ -523,6 +525,83 @@ Streaming.prototype.disposeSubscription = function(subscription) {
     }
 };
 
+function createHandleSubscriptionReadyForUnsubscribe(subscriptions, resolve) {
```
lukeapage: how about a top level function

handleSubscriptionReadyForUnsubscribe

taking the 2 arguments,

then to "create" the one to pass in, use bind

src/openapi/streaming/streaming.js
```diff
@@ -523,6 +525,83 @@ Streaming.prototype.disposeSubscription = function(subscription) {
...
+    return function() {
+        if (subscriptions.every((subscription) => subscription.isReadyForUnsubscribeByTag())) {
+            resolve();
+        }
+    };
+}
+
+function unsubscribeSubscriptionByTag(serviceGroup, url, tag, shouldDisposeSubscription) {
+    const subscriptionsToRemove = this.subscriptions.filter((subscription) =>
```
lukeapage: can we use .filter ? I can't remember what we say we support, did you check?

src/openapi/streaming/streaming.js
```diff
@@ -523,6 +525,83 @@ Streaming.prototype.disposeSubscription = function(subscription) {
...
+            subscription.subscriptionData.Tag === tag
+        );
+
+    let allSubscriptionsReadyResolve;
+    const allSubscriptionsReady = new Promise((resolve) => {
+        allSubscriptionsReadyResolve = resolve;
+    });
+    const handleSubscriptionReadyForUnsubscribe = createHandleSubscriptionReadyForUnsubscribe(subscriptionsToRemove, allSubscriptionsReadyResolve);
+
+    subscriptionsToRemove.forEach((subscription) => {
```
lukeapage: same here, the current "style" of this library is for loops rather than .forEeach

src/openapi/streaming/streaming.js
```diff
@@ -523,6 +525,83 @@ Streaming.prototype.disposeSubscription = function(subscription) {
...
+        this.transport.delete(serviceGroup, url + '/{contextId}/?Tag={tag}', {
+            contextId: this.contextId,
+            tag,
+        })
+        .then(() => subscriptionsToRemove.forEach((subscription) => {
+            subscription.onUnsubscribeByTagComplete();
+            subscription.removeStateChangedCallback(handleSubscriptionReadyForUnsubscribe);
+        }))
+        .catch(() => subscriptionsToRemove.forEach((subscription) => {
+            subscription.onUnsubscribeByTagComplete();
```
lukeapage: do we normally not do anything different if a unsubscribe call fails?
if we do keep this, do

.catch(noop)
.then(

to avoid duplicated code

src/openapi/streaming/subscription.js
```diff
@@ -163,6 +169,18 @@ function performAction(queuedAction) {
...
 
+        case ACTION_UNSUBSCRIBE_BY_TAG_PENDING:
+            switch (this.currentState) {
+                case STATE_SUBSCRIBED:
+                case STATE_UNSUBSCRIBED:
+                    unsubscribeByTagPending.call(this);
+                    break;
+
+                default:
+                    log.error(LOG_AREA, 'unanticipated state', { state: this.currentState, action });
```
lukeapage: why do we wait for subscriptions to be ready if er error unless the state is subscribed/unsubscirbed?

I was expecting that if you were unsubscribing, we ignore (double unsubscribe is no problem)
if you were subscribing, we wait until subscribed, and then change to ready for unsubscribe by tag

src/openapi/streaming/subscription.js
```diff
@@ -203,7 +221,7 @@ function onSubscribeSuccess(referenceId, result) {
         return;
     }
 
-    this.currentState = STATE_SUBSCRIBED;
```
lukeapage: we need unit tests - there should be pretty good coverage

src/openapi/streaming/subscription.js
jawalker88: Not sure I follow your question exactly but ... The only non-transitioning states are SUBSCRIBED and UNSUBSCRIBED, and we need to wait for all tagged subscriptions to be in one of these states. We need to handle both being subscribed and unsubscribed as some subscriptions could be unsubscribed but have the same tag as subscribed ones, we need to mark these as "ready" for unsubscribe by tag too.

I've added some logic in subscription-queue to handle unsubscribeByTag in a similar way to unsubscribe as I'd missed that.

lukeapage: I understand now

lukeapage:
LGTM. Will ask xmiku to take a look. is it ready to merge from your side?

jawalker88:
@lukeapage after Michal has reviewed, I think it's good from my side.
You mentioned about publishing a beta, could we do that while testing in case more changes are needed?

src/openapi/streaming/streaming.js
```diff
@@ -523,6 +525,103 @@ Streaming.prototype.disposeSubscription = function(subscription) {
...
+    }
+
+    return subscriptionsToRemove;
+}
+
+function unsubscribeSubscriptionByTag(serviceGroup, url, tag, shouldDisposeSubscription) {
+    const subscriptionsToRemove = getSubscriptionsToRemove.call(this, serviceGroup, url, tag);
+
+    let allSubscriptionsReadyResolve;
+    const allSubscriptionsReady = new Promise((resolve) => {
```
michax: What do you think about extracting Promise to separate function so We could end up with **unsubscribeSubscriptionByTag** looking like below. I think it would be easier to reason about.
```
function unsubscribeSubscriptionByTag(serviceGroup, url, tag, shouldDisposeSubscription) {
    const subscriptionsToRemove = getSubscriptionsToRemove.call(this, serviceGroup, url, tag);
    const allSubscriptionsReady = getSubscriptionsReadyPromise.call(this, subscriptionsToRemove, onStateChanged, shouldDisposeSubscription);

    allSubscriptionsReady.then(() => {
        ...
    });
}
```

src/openapi/streaming/streaming.js
```diff
@@ -523,6 +525,103 @@ Streaming.prototype.disposeSubscription = function(subscription) {
     }
 };
 
+function handleSubscriptionReadyForUnsubscribe(subscriptions, resolve) {
```
michax: Uber nitpick. What do you think about moving those new private functions above Streaming constructor. It seems like a general structure we use for this file (private methods and then constructor and public ones). It makes it more consistent, i think.

src/openapi/streaming/streaming.js
```diff
@@ -523,6 +525,103 @@ Streaming.prototype.disposeSubscription = function(subscription) {
...
+};
+
+/**
+ * Disposes all subscriptions at the given serviceGroup and url by tag permanently. They will be stopped and not be able to be started.
+ *
+ * @param {string} serviceGroup - the serviceGroup of the subscriptions to unsubscribe
+ * @param {string} url - the url of the subscriptions to unsubscribe
+ * @param {string} tag - the tag of the subscriptions to unsubscribe
+ */
+Streaming.prototype.disposeByTag = function(serviceGroup, url, tag) {
```
michax: Rename to _disposeSubscriptionByTag_, to be more specific, and more consistent with existing disposeSubscription which is similar, but for specific subscription instead of group of tagged ones.

src/openapi/streaming/streaming.js
```diff
@@ -523,6 +525,103 @@ Streaming.prototype.disposeSubscription = function(subscription) {
...
+            allSubscriptionsReady = false;
+        }
+    }
+
+    if (allSubscriptionsReady) {
+        resolve();
+    }
+}
+
+function getSubscriptionsToRemove(serviceGroup, url, tag) {
```
michax: Better name for this function would be getSubscriptionsByTag. It doesn't have any filtering logic that would indicate that  given subscription is pending for removal,  and that's what i though when i first read this initial name.

src/openapi/streaming/streaming.js
```diff
@@ -523,6 +525,103 @@ Streaming.prototype.disposeSubscription = function(subscription) {
...
+        const subscription = subscriptionsToRemove[i];
+
+        ignoreSubscriptions[subscription.referenceId] = true;
+        setTimeout(() => delete ignoreSubscriptions[subscription.referenceId], MS_TO_IGNORE_DATA_ON_UNSUBSCRIBED);
+
+        subscription.addStateChangedCallback(onStateChanged);
+        subscription.onUnsubscribeByTagPending();
+
+        if (shouldDisposeSubscription) {
+            subscription.dispose();
```
michax: Disposal logic is duplicate of one found inside Streaming.prototype.disposeSubscription.
Consider extracting common part, maybe as remoteSubscription, and use here and in Streaming.prototype.disposeSubscription. Should further improve readability of unsubscribeSubscriptionByTag.

src/openapi/streaming/subscription.js
```diff
@@ -394,6 +421,24 @@ Subscription.prototype.UPDATE_TYPE_DELTA = 2;
 Subscription.prototype.OPENAPI_DELETE_PROPERTY = '__meta_deleted';
 
 /**
+ * Add a callback to be invoked when the subscription state changes.
+ */
+Subscription.prototype.addStateChangedCallback = function(callback) {
+    this.onStateChangedCallbacks.push(callback);
```
michax: Before pushing, check if it already exists.

michax:
Generally looks good. I've added few suggestions to improve readability. Let me know what you think about them.

src/openapi/streaming/streaming.js
```diff
@@ -523,6 +525,103 @@ Streaming.prototype.disposeSubscription = function(subscription) {
...
+    const allSubscriptionsReady = new Promise((resolve) => {
+        allSubscriptionsReadyResolve = resolve;
+    });
+    const onStateChanged = handleSubscriptionReadyForUnsubscribe.bind(this, subscriptionsToRemove, allSubscriptionsReadyResolve);
+
+    for (let i = 0; i < subscriptionsToRemove.length; i++) {
+        const subscription = subscriptionsToRemove[i];
+
+        ignoreSubscriptions[subscription.referenceId] = true;
+        setTimeout(() => delete ignoreSubscriptions[subscription.referenceId], MS_TO_IGNORE_DATA_ON_UNSUBSCRIBED);
```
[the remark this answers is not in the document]
jawalker88: I think it should clear if unsubscribe is interrupted as we then shouldn't be ignoring the subscription any more

PR #55, opened 2017-11-08
Feature pagination

michax:
Add support for pagination, following requirements from backend openapi documentation.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,19 @@
+#### v1.9.1
+
++ Add nano-precision to protobuf deserialization
+
+#### v1.9.0
+
++ Add precision parameter to short formatter
+
+#### v1.8.0
+
++ Support for Protobuf deserialization
+
+#### v1.7.0
+
++ Add ability to unsubscribe by tag
+
 #### v1.6.0
 
 + No longer send /active as part of the subscribe url
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "openapi-clientlib",
-  "version": "1.8.1-pagination-beta",
+  "version": "1.9.2-pagination-beta",
   "engines": {
     "node": ">=4"
   },
--- a/src/number-formatting/number-formatting.js
+++ b/src/number-formatting/number-formatting.js
@@ -88,10 +88,11 @@ NumberFormatting.prototype.formatNoRounding = function(num, minDecimals, maxDeci
 /**
  * Formats a number into a short format, e.g. 10.000 becomes 10k.
  * @param {number} number
+ * @param {number} precision
  * @returns {string}
  */
-NumberFormatting.prototype.shortFormat = function(number) {
-    return shortFormat(number, this);
+NumberFormatting.prototype.shortFormat = function(number, precision) {
+    return shortFormat(number, precision, this);
 };
 
 /**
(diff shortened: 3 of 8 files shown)


src/openapi/streaming/subscription.js
```diff
@@ -35,14 +35,22 @@ const LOG_AREA = 'Subscription';
...
 // -- Local methods section --
 
-function getUrl(url, args) {
-    if (!args.Top) {
+function getUrl(url, subscriptionData) {
+    if (!subscriptionData.Top) {
         return url;
     }
 
     return url+'?$top={Top}';
```
michax: I don't like it. Strange idea to drive pagination through url encoded parameters. @lukeapage What do you think about it? Maybe it's worth pushing openapi to just allow for passing Top parameters as part of standard arguments, next to Format and RefreshRate?

src/openapi/streaming/subscription.js
```diff
@@ -35,6 +35,22 @@ const LOG_AREA = 'Subscription';
 
 // -- Local methods section --
 
+function getUrlTemplate(url, subscriptionData) {
+    if (!subscriptionData.Top) {
+        return url;
+    }
+
+    return url + '?$top={Top}';
```
michax: I don't like it. Strange idea to drive pagination through url encoded parameters. @lukeapage What do you think about it? Maybe it's worth pushing openapi to just allow for passing Top parameters as part of standard arguments, next to Format and RefreshRate?

src/openapi/streaming/subscription.js
```diff
@@ -55,7 +71,11 @@ function subscribe() {
 
     log.debug(LOG_AREA, 'starting..', { serviceGroup: this.serviceGroup, url: this.url });
     this.currentState = STATE_SUBSCRIBE_REQUESTED;
-    this.transport.post(this.serviceGroup, this.url, null, { body: data })
+
+    const urlTemplate = getUrlTemplate(this.url, this.subscriptionData);
+    const templateArgs = getTemplateArgs(this.subscriptionData);
+
+    this.transport.post(this.serviceGroup, urlTemplate, templateArgs, { body: data })
```
lukeapage: the url template and template args are really only there as helpers and to to allow clearer code..

In this case, I'm not sure we need it.. shall we just have a getUrl method?

src/openapi/streaming/subscription.js
```diff
@@ -35,6 +35,22 @@ const LOG_AREA = 'Subscription';
...
+function getUrlTemplate(url, subscriptionData) {
+    if (!subscriptionData.Top) {
+        return url;
+    }
+
+    return url + '?$top={Top}';
+}
+
+function getTemplateArgs(subscriptionData) {
+    if (!subscriptionData.Top) {
```
lukeapage: so I guess open api doesn't mind having the Top argument sent in its subscription data?

michax: I don't think it will affect anything but I will removed for clarity. Otherwise it can be misleading when evaluating requests.

src/openapi/streaming/subscription.js
michax: I'll need to check it. I assumed they also normalize,escape url.

michax:
@lukeapage Resolved your comments.

lukeapage:
Shall I publish this beta or did I already do it?

PR #58, opened 2018-01-24
handle  control message `_disconnect`

tushar-chandra-030389:
logic to handle new control message `_disconnect`


src/openapi/streaming/streaming.js
```diff
@@ -306,6 +311,17 @@ function resetSubscriptions(referenceIdList) {
     }
 }
 
+/**
+ * on disconnect control message, throw error in all subscriptions and dispose streaming
+ */
+function handleDisconnect() {
+    for(let i = 0, subscription; subscription = this.subscriptions[i]; i++) {
+      onSubscribeError.call(subscription, subscription.referenceId, OPENAPI_CONTROL_MESSAGE_DISCONNECT);
```
lukeapage: where is onSubscribeError defined?
I'd expect the open api client library to maybe tell subscriptions the connection is unavailable so they don't try and reconnect and then send an event that disconnect has been requested.

this allows consumers to either log out straight away or handle it by showing a message.

PR #60, opened 2018-02-14
adding check for negative zero

abpasaxo:
Adding check for '-0' in case of formatting infinitesimally small negative values. The output should be '0'

### DIFF
--- a/src/price-formatting/format.js
+++ b/src/price-formatting/format.js
@@ -199,7 +199,10 @@ function formatPriceParts(numberFormatting, value, decimals, formatFlags, numera
 
     if (isNegative) {
         parts.Post = numberFormatting.negativePost;
-        parts.Pre = numberFormatting.negativePre;
+
+        // Infinitesimally small negative value is rounded to 0, in which case Pre should not be '-'
+        // as '-0' makes no sense, hence the below check
+        parts.Pre = (parts.First === '0' && !parts.Pips) ? '' : numberFormatting.negativePre;
     }
 
     return parts;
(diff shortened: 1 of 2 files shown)


src/price-formatting/format.js
```diff
@@ -199,7 +199,10 @@ function formatPriceParts(numberFormatting, value, decimals, formatFlags, numera
 
     if (isNegative) {
         parts.Post = numberFormatting.negativePost;
-        parts.Pre = numberFormatting.negativePre;
+
+        // Infinitesimally small negative value is rounded to 0, in which case Pre should not be '-'
+        // as '-0' makes no sense, hence the below check
+        parts.Pre = (parts.First === '0' && !parts.Pips) ? '' : numberFormatting.negativePre;
```
lukeapage: To be thorough, should we check decipips too?

src/price-formatting/format.js
```diff
@@ -199,7 +199,10 @@ function formatPriceParts(numberFormatting, value, decimals, formatFlags, numera
 
     if (isNegative) {
         parts.Post = numberFormatting.negativePost;
```
lukeapage: the same check needs to happen on post - thats for cultures where the - goes after or -1 is represented as (1)

abpasaxo: Adding check for post as well.

src/price-formatting/format.js
abpasaxo: I discussed this with Mayank, decipips won't come untill and unless there is a value for pips. The decipips check would be redundant. Hence skipping it.

lukeapage: @abpasaxo I think mayank is wrong .. it’s probably never going to happen but let’s code for it in case

abpasaxo: Added check for decipips.

lukeapage:
superceded by https://github.com/SaxoBank/openapi-clientlib-js/pull/63

PR #61, opened 2018-02-22
 Add precision parameter to short formatter

oskarszura:
Add precision parameter to short formatter

In some cases, short formatted numbers require custom precision. Developers should be able to define, with what precision shortened number should be returned.

The new signature of `shortFormat ` should be compatible with previous usages. To guarantee that a check for the presence of `precision` parameter is done. If the parameter is not present, the function will calculate the precision in the old way.

### DIFF
--- a/src/number-formatting/short-format.js
+++ b/src/number-formatting/short-format.js
@@ -17,7 +17,7 @@ import formatNumber from './format';
  * @param options
  * @returns {string} Returns 0 when dates are equal. -1 when date1 less than date2. 1 when date1 greater than date2.
  */
-function shortFormat(num, options) {
+function shortFormat(num, precision, options) {
     const [digits] = String(num).split('.');
     let digitSize = digits.length;
     let boundary;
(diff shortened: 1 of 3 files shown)


src/number-formatting/short-format.js
```diff
@@ -17,7 +17,7 @@ import formatNumber from './format';
  * @param options
  * @returns {string} Returns 0 when dates are equal. -1 when date1 less than date2. 1 when date1 greater than date2.
  */
-function shortFormat(num, options) {
+function shortFormat(num, precision, options) {
```
michax: Add new parameter to JSDocs. Not only here but in other places where we added new argument.

src/number-formatting/short-format.js
```diff
@@ -14,10 +14,11 @@ import formatNumber from './format';
 /**
  * Converts from a number to a string like "1k" or "100m".
  * @param num
+ * @param precision
  * @param options
  * @returns {string} Returns 0 when dates are equal. -1 when date1 less than date2. 1 when date1 greater than date2.
  */
-function shortFormat(num, options) {
+function shortFormat(num, precision, options) {
```
michax: Add new parameter to JSDocs. Not only here but in other places where we added new argument.

PR #62, opened 2018-02-26
Adding nanoprecision to protobuf timestamp response

majtsaxo:
Relates to any timestamp given that should be comparable by nano precision.

### DIFF
--- a/src/openapi/streaming/serializer/wrappers/protobuf-wrappers.js
+++ b/src/openapi/streaming/serializer/wrappers/protobuf-wrappers.js
@@ -7,8 +7,10 @@ export default {
                 },
 
                 toObject(message, options) {
-                    const seconds = Number(message.seconds) * 1000;
-                    const date = new Date(seconds);
+                    const { seconds, nanos } = message;
+
+                    // Date with support for nano precision
+                    const date = new Date((Number(seconds) * 1000) + Math.floor((Number(nanos) / 1000000)));
 
                     return date.toJSON();
                 },
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "openapi-clientlib",
-  "version": "1.9.0",
+  "version": "1.9.0-protobuf-timestamp-nanoprecision",
   "engines": {
     "node": ">=4"
   },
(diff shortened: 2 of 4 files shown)


src/openapi/streaming/serializer/wrappers/protobuf-wrappers.js
```diff
@@ -7,10 +7,34 @@ export default {
...
+                    const NANOS_PER_MILLISECOND = 1000000;
+                    const NANOS_PER_MICROSECOND = 1000;
 
-                    return date.toJSON();
+                    const date = new Date(Number(seconds) * MILIS_PER_SECOND);
+
+                    let result = date.toJSON();
+
+                    // Support for Nano precision
+                    if (nanos !== 0) {
```
michax: Nice find. Thanks!

Could you add some test cases to serializer-protobuf-spec.js, with mocked nanoseconds in proto-price.js?

src/openapi/streaming/serializer/wrappers/protobuf-wrappers.js
```diff
@@ -7,10 +7,34 @@ export default {
                 },
 
                 toObject(message, options) {
-                    const seconds = Number(message.seconds) * 1000;
-                    const date = new Date(seconds);
+                    const { seconds, nanos } = message;
+                    const MILIS_PER_SECOND = 1000;
```
michax: Maybe we can simplify it. Check google timestamp wrappere for JS:
https://github.com/google/protobuf/blob/master/src/google/protobuf/compiler/js/well_known_types/timestamp.js

src/openapi/streaming/serializer/wrappers/protobuf-wrappers.js
majtsaxo: Yes, will do.

src/openapi/streaming/serializer/wrappers/protobuf-wrappers.js
majtsaxo: Good idea!

src/openapi/streaming/serializer/wrappers/protobuf-wrappers.js
```diff
@@ -7,8 +7,10 @@ export default {
                 },
 
                 toObject(message, options) {
-                    const seconds = Number(message.seconds) * 1000;
-                    const date = new Date(seconds);
+                    const { seconds, nanos } = message;
+
+                    // Date with support for nano precision
+                    const date = new Date((Number(seconds) * 1000) + (Number(nanos) / 1000000));
```
lukeapage: Shall we adjust this to be an integer? e.g.

Math.floor(Number(nanos) / 1000000)

so its clear this is to get the ms, not that Date supports fractions..

PR #64, opened 2018-03-20
Short formatter remove 'zero tails' by default

oskarszura:
Short formatter remove 'zero tails' by default

Short formatter should not allow manually set precision for formatted numbers but when shortened number has tail only from `0` digits (ex. `10.00k` / `50.0k` etc.) it should be removed (`10.00k` into `10k` / `50.0k` into `50k`).

The `expandNumber` function has a new option flag, which is turned on by default for `shortFormat` function (we want to keep the result of the short format as short as possible). This flag prevents displaying of the floating part (only when made of `0` digits) and unnecessary related calculations.

### DIFF
--- a/src/number-formatting/format.js
+++ b/src/number-formatting/format.js
@@ -30,10 +30,10 @@ function convertNumbertToString(number, precision) {
  * expands the number of decimals and introduces decimal groups.
  * @param number
  * @param precision
- * @param { groupSizes, sep, decimalChar, isHideZeroTail } options
+ * @param { groupSizes, groupSeparator, decimalSeparator, isHideZeroTail } options
  */
 function expandNumber(number, precision, options) {
-    const { groupSizes, sep, decimalChar, isHideZeroTail } = options;
+    const { groupSizes, groupSeparator, decimalSeparator, isHideZeroTail } = options;
     let curSize = groupSizes[0];
     let curGroupIndex = 1;
     let numberString = convertNumbertToString(number, precision);
@@ -60,7 +60,7 @@ function expandNumber(number, precision, options) {
             }
         }
 
-        right = decimalChar + right;
+        right = decimalSeparator + right;
     } else {
         right = '';
     }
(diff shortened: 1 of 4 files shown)


src/number-formatting/format.js
```diff
@@ -105,10 +107,12 @@ function formatNumber(inputNumber, decimals, options) {
     absoluteNumber = Math.round(absoluteNumber * factor) / factor;
 
     let formattedNumber = expandNumber(Math.abs(absoluteNumber),
-                                decimals,
-                                options.groupSizes,
-                                options.groupSeparator,
-                                options.decimalSeparator);
+                                decimals, {
```
michax: for readability I would move payload to separate const and use it as expandNumber argument.

```expandNumber(Math.abs(absoluteNumber, decimals, formatOptions);```

File: src/number-formatting/short-format.js [hunk 2]
```diff
@@ -18,7 +18,8 @@ import formatNumber from './format';
  * @param options
  * @returns {string} Returns 0 when dates are equal. -1 when date1 less than date2. 1 when date1 greater than date2.
  */
-function shortFormat(num, precision, options) {
+function shortFormat(num, options) {
+    const shortFormatOptions = $.extend({}, options, { isHideZeroTail: true });
```
michax: Consider using extend from utils/object

src/number-formatting/format.js
```diff
@@ -104,11 +106,14 @@ function formatNumber(inputNumber, decimals, options) {
...
     absoluteNumber = Math.round(absoluteNumber * factor) / factor;
 
-    let formattedNumber = expandNumber(Math.abs(absoluteNumber),
-                                decimals,
-                                options.groupSizes,
-                                options.groupSeparator,
-                                options.decimalSeparator);
+    const opts = {
+        groupSizes: options.groupSizes,
+        sep: options.groupSeparator,
```
lukeapage: We are breaking expand number contract anyway - why use different option names than the ones passed in?

I prefer the explicit groupSeparator to sep and decimalSeparator to decimalChar

PR #65, opened 2018-04-24
Handle negative input in short-format

sankalplakhina:


src/number-formatting/short-format.js
```diff
@@ -20,6 +20,12 @@ import formatNumber from './format';
  * @returns {string} Returns 0 when dates are equal. -1 when date1 less than date2. 1 when date1 greater than date2.
  */
 function shortFormat(num, options) {
+    let minusSign = '';
+    if (num < 0) {
+        minusSign = '-';
```
lukeapage: This isn't culturally correct - price formatting uses negativePre and negativePost
number format uses negativePattern

lukeapage: I assume we are doing some cultural things so that "," or "." is used correctly?

sankalplakhina: Got it! I have updated the code and added more test cases for to check for `negativePost` scenario.

PR #66, opened 2018-06-07
If there is a network disconnection, stop the transport from retrying

tushar-chandra-030389:
If there is a network disconnection, stop the transport from retrying

### DIFF
--- a/src/openapi/transport/retry.js
+++ b/src/openapi/transport/retry.js
@@ -48,8 +48,8 @@ function transportMethod(method) {
  * var transportRetry = new TransportRetry(transport, {
  *      retryTimeout:10000,
  *      methods:{
- *          'delete':{ retryLimit:3 },
- *          'post':{ retryTimeouts: [1000, 1000, 2000, 3000, 5000], statuses: [504] },
+ *          'delete':{ retryLimit:3, retryNetworkError: true },
+ *          'post':{ retryTimeouts: [1000, 1000, 2000, 3000, 5000], statuses: [504], retryNetworkError: false },
  *      }
  * });
  */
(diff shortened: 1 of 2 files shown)


src/openapi/transport/retry.js
```diff
@@ -128,7 +128,11 @@ TransportRetry.prototype.sendTransportCall = function(transportCall) {
                 const isRetryRequest = !(response && response.status) || callOptions.statuses && callOptions.statuses.indexOf(response.status) >= 0;
                 const isWithinRetryLimitOption = callOptions.retryLimit > 0 && transportCall.retryCount < callOptions.retryLimit;
                 const isWithinRetryTimeoutsOption = callOptions.retryTimeouts && transportCall.retryCount < callOptions.retryTimeouts.length;
-                if (isRetryRequest && (isWithinRetryLimitOption || isWithinRetryTimeoutsOption) && !this.isDisposed) {
+
+                if (!response.isNetworkError && isRetryRequest && (
```
tushar-chandra-030389: @lukeapage 

We should only stop the retry in case of network error, if the method type is POST,
else it should be added to a failed request?

src/utils/fetch.js
```diff
@@ -61,7 +61,12 @@ export function convertFetchResponse(url, body, result, isRejected) {
             result,
         });
 
-        throw result;
+        const networkError = {
+            message: result.message,
```
lukeapage: how do you know there is a result.message? on every browser? I suggest result || result.message

src/openapi/transport/retry.js
```diff
@@ -1,4 +1,4 @@
-﻿/**
```
lukeapage: have this added BOM's? or were BOM's there before

src/utils/fetch.js
```diff
@@ -1,4 +1,4 @@
-﻿/**
```
lukeapage: BOM's

src/openapi/transport/retry.js
lukeapage: Can I suggest this:

  *          'delete':{ retryLimit:3 },
  *          'post':{ retryTimeouts: [1000, 1000, 2000, 3000, 5000], statuses: [504] },

becomes

  *          'delete':{ retryLimit:3, retryNetworkError: true },
  *          'post':{ retryTimeouts: [1000, 1000, 2000, 3000, 5000], statuses: [504], retryNetworkError: false },

so we explicitly define in the options whether to retry a network request

src/openapi/transport/retry.js
tushar-chandra-030389: resolved.

src/utils/fetch.js
tushar-chandra-030389: resolved.

PR #67, opened 2018-06-11
Add short hand notation for billion

varnikaj:
Add short hand notation for billion


src/number-formatting/short-format.js
```diff
@@ -40,6 +40,11 @@ function shortFormat(num, options) {
         }
     }
 
+    if (digitSize >= 10) { // > 999,999,500
```
michax: Nitpic. In comment example we have 9 digit number. Is that correct?

varnikaj: Yes, it should be 9 digit number only, so any 9 digit number after mentioned number is rounded off to billion

lukeapage: I'm not sure the comments make sense..
surely

if (digitSize >= 10) { // >= 1,000,000,000

test/unit/number-formatting/format-spec.js
```diff
@@ -75,7 +75,7 @@ describe('NumberFormatting format', () => {
             expect(shortFormat(1050000)).toEqual('1.05m');
             expect(shortFormat(10500000)).toEqual('10.5m');
             expect(shortFormat(105000000)).toEqual('105m');
-            expect(shortFormat(1050000000)).toEqual('1,050m');
+            expect(shortFormat(1050000000)).toEqual('1.05bn');
```
lukeapage: I think you should add more than a single assertion. E.g. the boundaries on both sides you expect. a larger number. etc.

short-format.js
```diff
@@ -0,0 +1,100 @@
...
+
+// -- Exported methods section --
+
+/**
+ * Converts from a number to a string like "1k" or "100m".
+ * @param num
+ * @param precision
+ * @param options
+ * @returns {string} Returns 0 when dates are equal. -1 when date1 less than date2. 1 when date1 greater than date2.
+ */
```
varnikaj: Refactored code a bit

PR #68, opened 2018-06-27
Un-subscribe after network errors

tushar-chandra-030389:


src/openapi/streaming/subscription.js
```diff
@@ -288,6 +288,16 @@ function onSubscribeSuccess(referenceId, result) {
  * @param response
  */
 function onSubscribeError(referenceId, response) {
+    if (response.isNetworkError) {
+        this.onUnsubscribe();
+
+        if (this.onError) {
```
lukeapage: ah I see, it misses out
 if (this.queue.peekAction() !== ACTION_UNSUBSCRIBE) {
too.

what about calling onUnsubscribe at the end e.g.

if (response.isNetworkError) {
   this.onUnsubscribe();
} else {
}

tushar-chandra-030389: Done.

Need to move `setState.call(this, STATE_UNSUBSCRIBED);` as well.
But I am not sure why we are settings the state, can't we just call this.onUnsubscribe?

PR #70, opened 2018-07-03
Global Request Id for Batches

michax:
This changes introduces global request ids for batches.
Previously, each batch child item had local request id counted from 0.
In new version, we always use global, unique per session x-request-id.

When parsing, we use parent x-request-id as a reference to calculate local position of each 
child batch item.


src/openapi/transport/batch.js
```diff
@@ -67,6 +80,7 @@ function batchCallSuccess(callList, batchResult) {
  * @param {Array.<{method: string, args:Array}>} callList
  */
 function runBatchCall(serviceGroup, callList) {
+    const parentRequestId = getRequestId();
```
michax: Request id for container request that contains all child batched requests.
It's need to be requested before child requests are build to ensure that parent id has lower value then first child 
request. This ensures easier mapping to local ids when parsing response.

michax:
@lukeapage Can you review?

src/utils/request.js
```diff
@@ -0,0 +1,16 @@
+// counter used for identifying unique request and it will share among all transports
+let requestCounter = 0;
+
+function resetCounter() {
+    requestCounter = 0;
+}
+
+function getRequestId() {
+    return ++requestCounter;
```
michax: @lukeapage I'm using same logic for request counter increment as before, where first request id is 1. Do you remember if it was requirement or can we move to starting request ids from 0?

It would make some unit test cleaner, that's all.

lukeapage: I don't recall it being a requirement or why it should be from 1.

src/openapi/batch-util.js
```diff
@@ -25,10 +26,10 @@ const LOG_AREA = 'batch';
  * Parses the response from a batch call.
  * @name saxo.openapi.batchUtil.parse
  * @param {string} responseText
+ * @param {Number} parentRequestId - The parent request id. Used as a base for calculating batch items request ids.
```
lukeapage: wouldn't it be simpler to just always ask for a request id form the global request id counter?

lukeapage: its for parsing?

michax: It's just for parsing and deriving local index from global x-request-id of each sub response, which have strict relation to parent request-id. This ensures same order of batched responses for request items.

I've tried to follow requirement of previous implementation for safety. 

It might be worth simplifying it further, for example if we can get confirmation from openapi that they ensure correct response order to match request order. From what I've saw, it seems to be true. If that's the case, we might drop parentRequestId and x-request-id all together in this parser. Will talk to them about it. 

However, from the existing implementation, it looks like there was a reason for using x-request-id for index of requestData mapping. Which implies that order might change.

src/utils/request.js
michax: Ok, will confirm with openapi, and 0 is valid x-request-id, I'll change the implementation to use it.

src/openapi/batch-util.js
michax: > wouldn't it be simpler to just always ask for a request id form the global request id counter?

I do exactly that for building requests. Check ```build``` method in this file (batch-util). 
There is no additional complexity for building other then making sure we get x-request-id for parent batched request header before we build children.

PR #72, opened 2018-07-10
Clear pending reconnect timeouts

michax:
Clean pending reconnect timeouts.

There is not particular issue encountered so far, but it hanging timeouts might cause unexpected hard 
to reproduce issues.


lukeapage:
@michax any plans to get this merged or should we abandon?

PR #73, opened 2018-07-25
Fix: Parse - Modern fractional negative price gets converted to positive

grajiv:


src/price-formatting/parse.js
```diff
@@ -96,84 +96,100 @@ function parseDecimalprice(numberFormatting, s, formatFlags) {
...
-                const pipPart = numberFormatting.parse(s.substring(pipIndex + 1).trim());
+        if (pipIndex + 1 < s.length) {
+            const pipPart = numberFormatting.parse(s.substring(pipIndex + 1).trim());
 
-                if (pipPart < denominator) {
-                    result = result += (pipPart / denominator);
+            if (pipPart < denominator) {
+                // checking of "-0" in integerPart
+                if (integerPart.length === 2 &&
+                    integerPart.substr(0, 1) === '-' &&
```
michax: Why not have explicit check for "-0" ie.
integerPart.substr(0, 2) === '-0'

src/price-formatting/parse.js
```diff
@@ -96,84 +96,100 @@ function parseDecimalprice(numberFormatting, s, formatFlags) {
...
+                    integerPart.substring(1) === '0') {
+                    result = -1 * (pipPart / denominator);
                 } else {
-                    result = NaN;
+                    result = result += (pipPart / denominator);
```
michax: I think it can just be:
```result += (pipPart / denominator);```

No need for ```result = ```

File: src/price-formatting/parse.js (hunk 1 above)
Comment (grajiv) [reply src/price-formatting/parse.js]: Made this changes -(integerPart && integerPart === '-0')

File: src/price-formatting/parse.js (hunk 2 above)
Comment (grajiv) [reply src/price-formatting/parse.js]: Done

File: src/price-formatting/parse.js [hunk 3]
```diff
@@ -96,84 +96,98 @@ function parseDecimalprice(numberFormatting, s, formatFlags) {
...
 
-                if (pipPart < denominator) {
-                    result = result += (pipPart / denominator);
+            if (pipPart < denominator) {
+                // checking of "-0" in integerPart
+                if (integerPart && integerPart === '-0') {
+                    result = -1 * (pipPart / denominator);
                 } else {
-                    result = NaN;
+                    result += (pipPart / denominator);
```
lukeapage: Is this correct for numbers below -1?
It seems it would add the fraction onto the negative number. It should determine sign and the either add or subtract the fractional part shouldn’t it?

Also look at negative detection- it’s more complex than just checking -0

grajiv: I do not see difference in the result if I changed from 
result=result += (pipPart / denominator); 
to 
result += (pipPart / denominator);
I am getting the same result. Can you please give any example in which you have doubt. I will test it. BTW I tested both the scenarios in which result will be negative and positive.

grajiv: If we test the value less than -1, test case is failed in the old code and changed code. It means that there is no relation with the changes. Not sure this function is developed considering the least value of -1. 
I checked with VALI to get any Future Strategy instrument whose normal price is less than -1. Unfortunately, there is no such instrument. I think less than -1 is not real scenario. Can you please suggest me any such instrument if you know

lukeapage:
Negative numbers are broken before you are right. I don’t want to fix small negative numbers but the more general case of all negative numbers, whether they currently exist or not.

src/price-formatting/parse.js
grajiv: There are 2 bugs reported for this. This logic runs in other applications as it is copied and it might be changed when we get valid prices. I think we can create technical enhancement to handle the prices. Please suggest.

lukeapage: I don't want to apply a partial fix thats a bit of a hack and obscure the real problem that negative numbers don't format properly..

grajiv: If we know the prices which is less than -1 and its expected format, its decimals and numerator decimals, then we can make general fix. Please suggest for the solution.

lukeapage: You can add unit tests and fix the code based on the tests.

As I said 1. you should detect negative using the same code as for integers

Once you have the sign you can add or subtract correctly.

michax: I agree with @lukeapage, We need to fix it for all negative values.

I think it's can be easily don, if we just detect that integer is negative and then instead of adding fraction we subtract it.

My change against master that worked with both **-0.5625** and **-2567.21875**

```
if (pipIndex !== -1) {
            const integerPart = s.substring(0, pipIndex).trim();
            const isNegative = integerPart.indexOf('-') === 0;

            if (integerPart.length > 0) {
                result = numberFormatting.parse(integerPart);
            } else {
                result = 0;
            }

            if (pipIndex + 1 < s.length) {
                const pipPart = numberFormatting.parse(s.substring(pipIndex + 1).trim());

                if (pipPart < denominator) {
                    if (isNegative) {
                        result -= (pipPart / denominator);
                    } else {
                        result += (pipPart / denominator);
                    }
                } else {
                    result = NaN;
                }
            }
        } else {
            result = numberFormatting.parse(s);
        }
```

michax: @lukeapage Hehe, looks like I've posted to late. But it seems we are on the same page here.

grajiv: Done.

grajiv: @lukeapage @michax  I checked the new values with Valentin. He is ok with these values. So can you please take a look on this PR again

src/price-formatting/parse.js
```diff
@@ -96,84 +96,99 @@ function parseDecimalprice(numberFormatting, s, formatFlags) {
...
     let result;
+    const separator = getModernFractionsSeparator(numberFormatting);
+    const denominator = 1 << decimals;
+
+    const pipIndex = s.indexOf(separator);
+    if (pipIndex !== -1) {
+        const integerPart = s.substring(0, pipIndex).trim();
+        let isNegative = false;
+        if (integerPart.length > 0) {
+            result = numberFormatting.parse(integerPart);
```
lukeapage: see number parse and I'd suggest doing this on the whole string e.g.

const signInfo = parseNumberNegativePattern(value, options, true);
    let sign = signInfo[0];
    const num = signInfo[1];
    if (sign === '') {
        sign = '+';
    }

this matches what we do in formatPriceParts where we add negative pre and negative post

src/price-formatting/parse.js
```diff
@@ -96,84 +97,100 @@ function parseDecimalprice(numberFormatting, s, formatFlags) {
...
 
-    if (formatFlags.ModernFractions) {
-        // special futures
+        const signInfo = parseNumberNegativePattern(integerPart, numberFormatting);
+        const isNegative = signInfo[0] === '-';
 
-        const separator = getModernFractionsSeparator(numberFormatting);
-        const denominator = 1 << decimals;
+        if (pipIndex + 1 < s.length) {
+            const pipPart = numberFormatting.parse(s.substring(pipIndex + 1).trim());
```
lukeapage: use signInfo[1]

lukeapage: maybe even just re-assign 's'
s = signInfo[1]

test/unit/price-formatting/parse-spec.js
```diff
@@ -69,6 +69,10 @@ describe('price-formatting parse', () => {
         testConversion(0, 5, priceFormatOptions.ModernFractions, 1);
 
         testConversion(0, 5, priceFormatOptions.ModernFractions, 2);
+
+        testConversion(-0.5625, 5, priceFormatOptions.ModernFractions);
+
+        testConversion(-17.25, 5, priceFormatOptions.ModernFractions);
```
michax: Nitpick. 

We might get additional coverage for free by duplicating previous cases with negative value as input:
testConversion(2567.90625, 5, priceFormatOptions.ModernFractions);
```
        testConversion(-2567.21875, 5, priceFormatOptions.ModernFractions);
        testConversion(-2567.21875, 5, priceFormatOptions.ModernFractions, 1);
        testConversion(-2567.21875, 5, priceFormatOptions.ModernFractions, 2);
        testConversion(-2567.90625, 5, priceFormatOptions.ModernFractions, 1);
        testConversion(-2567.90625, 5, priceFormatOptions.ModernFractions, 2);
        testConversion(-2567.9921875, 5, priceFormatOptions.ModernFractions, 2);
```

grajiv: Done

PR #75, opened 2018-08-12
A few tweaks to better detect subscription problems

lukeapage:


src/openapi/streaming/subscription.js
```diff
@@ -19,7 +19,7 @@ import SerializerFacade from './serializer-facade';
 /**
  * The static counter to generate unique reference id's.
  */
-let referenceIdCounter = 0;
+let referenceIdCounter = 1;
```
lukeapage: should make no difference but this is just in case since the problems occur on subscription id 0

michax: @lukeapage It would be good to confirm with openapi that they don't rely on ReferenceId starting from 0.

PR #76, opened 2018-08-12
improve logging and functionality when a auth token is expired

lukeapage:


src/openapi/transport/auth.js
```diff
@@ -287,32 +312,6 @@ TransportAuth.prototype.refreshOpenApiToken = function() {
     }
 };
 
-/**
- * Tells transport auth that the token is invalid and it should immediately try to refresh if it isn't already
- */
-TransportAuth.prototype.onTokenInvalid = function() {
-
-    if (this.state & STATE_WAITING && this.retries === 0 && this.maxRetryCount > 0) {
```
michax: This logic wasn't moved to new implementation. Is it on purpose? 

Same question goes for those timeouts with margins:
```this.tokenRefreshTimerFireTime && (lateBy > (1000 * 20))```

PR #77, opened 2018-08-23
Feature - format swap prices

jwalker-scottlogic:
New price formatting needs to be added to support the new FX Swap price format. This needs to allow an additional decimal pip to be supplied and, optionally, the price to be formatted converted to pips. The changes to support this solution is as follows:

Add new options for specifying number of decimalPips and flag for formatAsPips.

Extend existing decipips logic to handle different decimalPips values.

When formatting as pips:
Round the price to Decimals + DecimalPips decimal places.
Remove the final DecimalPips digits as the DeciPips part.
Convert the remaining price to Pips as price * (10^Decimals)

### DIFF
--- a/test/unit/price-formatting/format-spec.js
+++ b/test/unit/price-formatting/format-spec.js
@@ -902,6 +902,28 @@ describe('price-formatting format', () => {
         expect(parts.Pips).toEqual('12,340');
         expect(parts.DeciPips).toEqual('.00');
         expect(parts.Post).toEqual('-');
+
+        // When FormatAsPips and decipips is a half
+        parts = priceFormatting.formatPriceParts(0.18885, 4, [priceFormatOptions.AllowDecimalPips, priceFormatOptions.FormatAsPips]);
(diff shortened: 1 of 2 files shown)


test/unit/price-formatting/format-spec.js
```diff
@@ -676,6 +770,64 @@ describe('price-formatting format', () => {
...
+        expect(parts.Post).toEqual('');
+
+        parts = priceFormatting.formatPriceParts(0.000001, 4, priceFormatOptions.AllowDecimalPips, null, 2);
+        expect(parts.Pre).toEqual('');
+        expect(parts.First).toEqual('0.00');
+        expect(parts.Pips).toEqual('00');
+        expect(parts.DeciPips).toEqual('01');
+        expect(parts.Post).toEqual('');
+
+        // When formatAsPips
```
mickylad: Does negative work too?

jawalker88: Added a couple of tests to cover negative prices

PR #79, opened 2018-08-24
Fix #78 - use default parameters instead of falsy check in formatNoRounding

### LINKED ISSUE #78: FormatNoRounding ingnores maxDecimals if it is set to 0
If the maxDecimals argument of `NumberFormatting.formatNoRounding` is passed as 0, it will be ignored because of the following line: https://github.com/SaxoBank/openapi-clientlib-js/blob/ac3072755b1b6a63b6d3aa7760c722b897a1fa2a/src/number-formatting/number-formatting.js#L79

grlira:


test/unit/number-formatting/format-spec.js
```diff
@@ -40,6 +40,11 @@ describe('NumberFormatting format', () => {
             expect(formatNumberNoRounding(-1.126, 1, 2)).toEqual('-1.13');
             expect(formatNumberNoRounding(-1.1, 4, 5)).toEqual('-1.1000');
             expect(formatNumberNoRounding(-5e-7, 4, 8)).toEqual('-0.0000005');
+
+            expect(formatNumberNoRounding(-1.1212, 1, 0)).toEqual('-1');
+            expect(formatNumberNoRounding(-1.126, 1, 0)).toEqual('-1');
+            expect(formatNumberNoRounding(-1.1, 4, 0)).toEqual('-1');
+            expect(formatNumberNoRounding(-5e-7, 4, 0)).toEqual('0');
```
raulpineda: is `formatNumberNoRounding(1.12345678, null, 4)` a valid use case? as in "I don't care what's the minimum amount of decimals as long as they aren't more than 4". If it is, right now it would equal `1` due to the parameter default, right?

raulpineda: Nevermind, I'm dumb 🤦‍♂️  a couple unit tests would be nice though :)

src/number-formatting/number-formatting.js
```diff
@@ -72,14 +72,7 @@ NumberFormatting.prototype.format = function(num, decimals) {
...
  */
-NumberFormatting.prototype.formatNoRounding = function(num, minDecimals, maxDecimals) {
-    if (!minDecimals) {
-        minDecimals = 0;
-    }
-    if (!maxDecimals) {
-        maxDecimals = 8;
-    }
-
+NumberFormatting.prototype.formatNoRounding = function(num, minDecimals = 0, maxDecimals = 8) {
```
michax: Please make sure We don't have existing behaviour which might rely on old logic ie. passing of null/NaN as maxDecimals. 

If we are not sure, it would be safer to have uglier option of explicit check against null, undefined and NaN.

lukeapage:
@grlira whats up with this PR - should we abandon it?

PR #80, opened 2018-09-24
on invalid batch responses, record the actual error

lukeapage:


src/openapi/transport/batch.js
```diff
@@ -56,7 +56,13 @@ function getParentRequestId(batchResult) {
 function batchCallSuccess(callList, batchResult) {
     const parentRequestId = getParentRequestId(batchResult);
```
michax: As micro improvement, You can move ```const parentRequestId = getParentRequestId(batchResult);```
below validation check.

src/openapi/transport/batch.js
```diff
@@ -56,7 +56,13 @@ function getParentRequestId(batchResult) {
 function batchCallSuccess(callList, batchResult) {
     const parentRequestId = getParentRequestId(batchResult);
 
+    if (!batchResult.response) {
```
michax: Would be good to have unit test for this part.

PR #82, opened 2018-10-10
log warnings when open api seems to take a long time

lukeapage:


src/utils/fetch.js
```diff
@@ -50,7 +52,9 @@ function convertFetchReject(url, body, error) {
  * @param result
  * @returns {Promise}
  */
-export function convertFetchSuccess(url, body, result) {
+export function convertFetchSuccess(url, body, timerId, result) {
```
michax: It looks like test use old signature and are failing with empty result. Please update tests.

PR #83, opened 2018-10-18
Fallback support for protobuf requests.

michax:
### Problem

Some openapi endpoints don't support protobuf (at this point only one).

### Solution

If protobuf is not supported by specific endpoint, try to subscribe using default json format.
Do it only once per subscription instance.

### DIFF
--- a/src/openapi/streaming/subscription.js
+++ b/src/openapi/streaming/subscription.js
@@ -34,6 +34,9 @@ const DEFAULT_REFRESH_RATE_MS = 1000;
 const MIN_REFRESH_RATE_MS = 100;
 
 const FORMAT_PROTOBUF = 'application/x-protobuf';
+const FORMAT_JSON = 'application/json';
+
+const ERROR_UNSUPPORTED_FORMAT = 'UnsupportedSubscriptionFormat';
 
 const LOG_AREA = 'Subscription';
 
@@ -287,7 +290,7 @@ function onSubscribeSuccess(referenceId, result) {
  * Called when a subscribe errors
  * @param response
  */
-function onSubscribeError(referenceId, response) {
+function onSubscribeError(referenceId, body) {
     if (referenceId !== this.referenceId) {
         log.debug(LOG_AREA, 'Received an error response for subscribing a subscription that has afterwards been reset - ignoring');
         return;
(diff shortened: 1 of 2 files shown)


src/openapi/streaming/subscription.js
```diff
@@ -441,6 +457,12 @@ function Subscription(streamingContextId, transport, serviceGroup, url, subscrip
     }
     this.connectionAvailable = true;
 
+    // Defines current number of fallback requests done due to invalid format.
+    this.formatFallbackCount = 0;
+
+    // Defines maximums fallback requests possible for invalid format.
+    this.maxFormatFallbacks = 1;
```
michax: At the moment don't see a reason for falling back more then once as part of specific subscription instance. 
Maybe instead of counters, simple flag in such case would be enough?

Need to also check if modify can bring protobuf format once again. AFAIK, it shouldn't (should only override part of arguments). However it might by that GO store will still hold protobuf format and pass it over on modify. Will validate such case.

src/openapi/streaming/subscription.js
```diff
@@ -287,15 +290,15 @@ function onSubscribeSuccess(referenceId, result) {
  * Called when a subscribe errors
  * @param response
  */
-function onSubscribeError(referenceId, response) {
+function onSubscribeError(referenceId, body) {
```
lukeapage: I don't think this is body - its an object
```
{
    response: {},
    status: 404,
}
```
I know its ugly doing response.response - thats historical. I think I wanted 
```
{
  body,
  status,
}
```
being the response

src/openapi/streaming/subscription.js
```diff
@@ -287,15 +290,15 @@ function onSubscribeSuccess(referenceId, result) {
...
     if (referenceId !== this.referenceId) {
         log.debug(LOG_AREA, 'Received an error response for subscribing a subscription that has afterwards been reset - ignoring');
         return;
     }
 
     setState.call(this, STATE_UNSUBSCRIBED);
     log.error(LOG_AREA, 'An error occurred subscribing', {
-        response,
+        body,
```
lukeapage: again, response is used by lots of log messages, I'd rtaher not add a variation

src/openapi/streaming/subscription.js
lukeapage: I'm not sure we need anything?
We already explicitly only fallback if the format is protobuf and we fallback to json, so isn't that enough?

michax: Should be enough. Just wanted to guard against openapi issue where they would wrongly return UnsupportedSubscriptionFormat error. We had issues in auth, which spammed us with retries due to invalid constant auth errors from openapi. 

Howevere if you believe that in this case it's to much, i can drop it. It's for sure less probable issue to happen then what we encountered with auth retry loop.

PR #84, opened 2018-11-08
Add authorization error (401) limits to block infinite loop of auth token requests

michax:
Fix introduces guard for maximum authorization fails that can happen within defined period of time (with debounce).

So for example if specific authorization endpoint fails 3 times, where each fail was within 5s timeframe, we block further error handling for this endpoint, until more then 5s passes.
This blocks further loop of endpoint errors and oauth re-requests.

Generally, scenario being fixed here is edge case end should never happen by design. However we had cases where openapi errors caused specific endpoint to always return 401 when requested.
By desing, on our side, if authorization error is returned for specific endpoint (which my happen by design), we request fresh authorization token. However, if endpoint is broken, and returns 401 for valid, fresh authorization tokens, we were re-requesting tokesn forever. And this change guards against infinite requests.

### DIFF
--- a/src/openapi/transport/auth.js
+++ b/src/openapi/transport/auth.js
@@ -22,8 +22,10 @@ const DEFAULT_TOKEN_REFRESH_CREDDENTIALS = 'include';
 
 // Max request limits are used to block infinite loop of authorization requests after transport 401 errors, which my happen if given
 // endpoint for whatever reasons constantly returns 401 error status (despite correct fresh authorization token refresh).
-const DEFAULT_MAX_AUTH_ERRORS = 10;
-const DEFAULT_AUTH_ERRORS_CLEANUP_DEBOUNCE = 1000;
+const DEFAULT_MAX_AUTH_ERRORS = 3;
+
+// Debounce time in milliseconds.
+const DEFAULT_AUTH_ERRORS_CLEANUP_DEBOUNCE = 5000; // ms
 
 const TOKEN_BEARER = 'Bearer ';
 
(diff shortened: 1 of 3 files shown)


src/openapi/transport/auth.js
```diff
@@ -20,6 +20,11 @@ const DEFAULT_TOKEN_REFRESH_PROPERTY_NAME_EXPIRES = 'expiry';
 const DEFAULT_TOKEN_REFRESH_MARGIN_MS = 0;
 const DEFAULT_TOKEN_REFRESH_CREDDENTIALS = 'include';
 
+// Max request limits are used to block infinite loop of authorization requests after transport 401 errors, which my happen if given
+// endpoint for whatever reasons constantly returns 401 error status (despite correct fresh authorization token refresh).
+const DEFAULT_MAX_AUTH_ERRORS = 10;
```
michax: Add comment describing what specificly those defaults drive.  Probably could just copy jsdocs for related parameters.

michax: It would be also good to rename those consts and related variable to make clear distinction between failing auth request and failing specific endpoints with auth status code.

src/openapi/transport/auth.js
```diff
@@ -220,6 +241,9 @@ function onTransportError(oldTokenExpiry, result) {
  * @param {string|number} [options.expiry] - The expiry of that token, assumed to be absolute.
  * @param {number} [options.retryDelayMs] - The delay before retrying auth
  * @param {number} [options.maxRetryCount] - The maximum number of times to retry the auth url
+ * @param {number} [options.maxAuthErrors] - The maximum number of authorization errors that
+ *          can occur for specific endpoint within specific timeframe.
+ * @param {number} [options.authErrorsCleanupDebounce] - The debounce timeout used for clearing of authorization errors count.
```
jakubsadura: nit: can you add a note that this is in ms?

PR #86, opened 2018-11-29
Remove grunt and upgrade dependencies

lukeapage:
Dependency on landing https://github.com/SaxoBank/openapi-clientlib-js/pull/85 first


.travis.yml
```diff
@@ -1,6 +1,3 @@
 language: node_js
-node_js: 
-  - "6"
-
-before_script:
-  - npm install -g grunt-cli
+node_js:
+  - "8"
```
jumoel: How about also specifying 10 (and perhaps 11?)

jumoel:
Have you considered switching to `yarn` and using regular SemVer ranges in `package.json`?

lukeapage:
> Have you considered switching to `yarn` and/or using regular SemVer ranges in `package.json`?

they are all dev dependencies.. so this stops random upgrades that could break things. IMO dev dependencies should always be exact and if your a package thats consumed, dependencies should be ranges. I've been burnt too many times by minor updates that break things.

.travis.yml
lukeapage: Changed to 10. Its only browser functionality, so I'm okay just testing in the latest.

jumoel:
> so this stops random upgrades that could break things.
That's why I suggest using Yarn. If `yarn.lock` hasn't been updated, it won't install other versions. On the CI, it could be run with `yarn install yarn install --frozen-lockfile` so it fails if `package.json` has been changed without `yarn.lock`.

lukeapage:
> > so this stops random upgrades that could break things.
> > That's why I suggest using Yarn. If `yarn.lock` hasn't been updated, it won't install other versions. On the CI, it could be run with `yarn install yarn install --frozen-lockfile` so it fails if `package.json` has been changed without `yarn.lock`.

Okay, sounds better. But I'm not going to switch in this PR. Maybe we can switch saxotrader at the same time so everything is aligned.

jumoel:
Fine with me 🎉

PR #87, opened 2019-01-08
feat: add Request ID from OpenAPI to error logging

jumoel:
Adds request ID passed from OpenAPI to the error logging.

Note: The header name might change after discussion with the OpenAPI teams.

Internal PR #: 30747

### DIFF
--- a/src/utils/fetch.js
+++ b/src/utils/fetch.js
@@ -137,7 +137,7 @@ export function convertFetchSuccess(url, body, timerId, result) {
             const correlation = result.headers.get('x-correlation') || '';
             
             // Form of correlation header is: {sessionId}#{AppId}#{requestId}#{serverDigits}
-            const [,,requestId] = correlation.split('#'); 
+            const requestId = correlation.split('#')[2]; 
 
             log.error(LOG_AREA, 'rejected server response', {
                 url,
--- a/src/utils/fetch.js
+++ b/src/utils/fetch.js
@@ -135,9 +135,9 @@ export function convertFetchSuccess(url, body, timerId, result) {
     if ((result.status < 200 || result.status > 299) && result.status !== 304) {
         convertedPromise = convertedPromise.then((newResult) => {
             const correlation = result.headers.get('x-correlation') || '';
-            
+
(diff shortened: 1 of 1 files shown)


src/utils/fetch.js
```diff
@@ -134,11 +134,13 @@ export function convertFetchSuccess(url, body, timerId, result) {
 
     if ((result.status < 200 || result.status > 299) && result.status !== 304) {
         convertedPromise = convertedPromise.then((newResult) => {
+            const requestId = parseInt(result.headers.get('x-request-id'), 10);
             log.error(LOG_AREA, 'rejected server response', {
                 url,
                 body,
                 status: newResult.status,
                 response: newResult.response,
+                requestId: isNaN(requestId) ? null : requestId,
```
michax: Nitpick. 
What do you think about not including **requestId** at all if it's not part of a header or is nan? 
Will make a error logging payload a little bit smaller.

michax:
Dev tested and reviewed. Looks/works fine.

src/utils/fetch.js
jumoel: Does the existence of the field (even if empty) make a deep dive through logs easier? I haven't had the need yet, so I don't know the possible shortcomings of our tools :)

My thought process was to aid in inspection capability (because that's what logs are for), but if theres no difference, then performance obviously wins.

michax: Yeah, null might be beneficial to detect that we indeed report it but it was empty. Otherwise there will be distinction between old code version which didn't include it at all. 

Let's leave it.

jumoel:
@michax @lukeapage How do you feel about merging this now?

It means the internal PR can be tested and perhaps merged. The result would be empty request IDs in the logs, until the API changes have been deployed.

michax:
> @michax @lukeapage How do you feel about merging this now?
> 
> It means the internal PR can be tested and perhaps merged. The result would be empty request IDs in the logs, until the API changes have been deployed.

@jumoel Will do some quick dev testing for exceptions. Code looks fine from my side. If no objections from @lukeapage I can merge it and prepare release build for you after dev testing.

PR #89, opened 2019-01-08
log when streaming reconnects using longpolling

staunsholm:
We want to monitor if and when streaming falls back to long polling.

This adds a simple log.warn when reconnecting using longPolling (usually after trying webSockets first).


src/openapi/streaming/streaming.js
```diff
@@ -148,6 +148,9 @@ function onConnectionStateChanged(change) {
             break;
 
         case this.CONNECTION_STATE_RECONNECTING:
+            if (signalRTransport && signalRTransport.name === 'longPolling') {
+                log.warn(LOG_AREA, 'reconnecting using longPolling');
```
michax: If nothing changed as of late, We only persist by default ERROR logs. I assume this logging was mostly intended as perisisted log to check stats for fallbacks to long polling among users.

src/openapi/streaming/streaming.js
```diff
@@ -148,6 +148,9 @@ function onConnectionStateChanged(change) {
             break;
 
         case this.CONNECTION_STATE_RECONNECTING:
+            if (signalRTransport && signalRTransport.name === 'longPolling') {
```
michax: There is one issue with this approach. It will log also all reconnect occurrences withing longPolling session. 
Are we interested in those as well or only scenarios where we transition from webSocket to longPolling transport?

src/openapi/streaming/streaming.js
staunsholm: added an isPersist flag (which the logger must handle)

src/openapi/streaming/streaming.js
staunsholm: Now only logs when transport changes (so no initial or reconnecting attempts).

src/openapi/streaming/streaming.js
```diff
@@ -150,10 +150,16 @@ function onConnectionStateChanged(change) {
             break;
 
         case this.CONNECTION_STATE_RECONNECTING:
+            // logs when longPolling is enabled after trying webSockets
+            if (signalRTransport && signalRTransport !== this.currentTransport && signalRTransport.name === 'longPolling') {
+                log.warn(LOG_AREA, 'changing transport to long polling', true);
```
michax: Isn't 3rd argument used for additional data? And 4th one for enabling persisting of warn?

staunsholm: It is...

PR #91, opened 2019-01-14
Default body to empty object for PATCH requests without body.

michax:
Change: Default body to empty object for PATCH requests without body

Reason: Some proxies and default configs for CDNs like Akamai have issues with accepting PATCH with content-length: 0.

### DIFF
--- a/src/utils/fetch.js
+++ b/src/utils/fetch.js
@@ -149,14 +149,14 @@ export function convertFetchSuccess(url, body, timerId, result) {
 }
 
 function getBody(method, options) {
-    if (!options) {
-        return null;
-    }
-
     // If PATCH without body occurs, create empty payload.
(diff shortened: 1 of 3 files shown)


src/utils/fetch.js
```diff
@@ -148,6 +148,20 @@ export function convertFetchSuccess(url, body, timerId, result) {
...
 
+function getBody(method, options) {
+    if (!options) {
+        return null;
+    }
+
+    // If PATCH without body occurs, create empty payload.
+    // Reason: Some proxies and default configs for CDNs like Akamai have issues with accepting PATCH with content-length: 0.
+    if (method === 'PATCH' && !options.body && !options.useXHttpMethodOverride) {
+        return {};
```
michax: Empty object seems like safest option here. It's what openapi does in explorer. This also enforcers json format, which some endpoints expect. 

Will syncup with openapi if something might potentially break if we send just string with empty space.

However at this point this is the safest option.

PR #92, opened 2019-02-21
Bug protobuf fallback logic

michax:
Fix for protobuf logic.

## Problem
Currently when endpoint doesn't support protobuf, endpoint returns an error and openapi-clientlib fallbacks to json subscription.
However in addition to a subscribe, we also send an error message which in our GO sagas drops/locks subscription from further passing updates.

## Solution
To be backward compatible, I've dropped invoking error callbacks when endpoint fails due to protobuf format being unsupported.
This ensure that userspace can keep current error handling logic with duplicated code need for handling protobuf errors.
So in short. Due to error being inreversible/fatal one, we only now send it whenever subscription fails without fallback. However in fallback scenarios fatal error isn't sent.

## Alternative options

If we decide to keep error being invoked, we should then drop re-connect logic as well and require it to be implemented on user space side.

### DIFF
--- a/src/openapi/streaming/subscription.spec.js
+++ b/src/openapi/streaming/subscription.spec.js
@@ -154,6 +154,7 @@ describe('openapi StreamingSubscription', () => {
             transport.post.mockClear();
 
             setTimeout(() => {
+                expect(errorSpy.mock.calls.length).toEqual(0);
                 expect(subscription.subscriptionData.Format).toEqual('application/json');
 
                 expect(transport.post.mock.calls.length).toEqual(1);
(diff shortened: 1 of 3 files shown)


src/openapi/streaming/subscription.spec.js
```diff
@@ -154,8 +154,6 @@ describe('openapi StreamingSubscription', () => {
             transport.post.mockClear();
 
             setTimeout(() => {
-                expect(errorSpy.mock.calls.length).toEqual(1);
```
lukeapage: we should test its *not* called?

michax: @lukeapage Good point. Done.

PR #94, opened 2019-03-11
Bug - do not attempt to stringify FormData

ToreJuloe:
Checking whether the data should needs to be `JSON.stringify`ed  by doing `typeof body === 'object'` means, that objects that instances of `Blob`, `ArrayBuffer`, `FormData`, `URLSearchParams` and `ReadableStream` will yield empty objects, instead of being handled properly.

This bad assumption breaks the ability to upload files, since we would have to rely on the `FileReader` API and base64 encode, which, aside from the overhead of doing this, may lead to unwanted behaviour due to arbitrary file size limits in different browsers.


src/utils/fetch.js
```diff
@@ -237,6 +237,17 @@ function localFetch(method, url, options) {
         );
 }
 
+// Check for handled type: https://fetch.spec.whatwg.org/#bodyinit
+// URLSearchParams and ReadableStream are guarded, because they are not supported in IE
+// USVString is not handled because it will be typeof "string"
+function isAlreadySupported(body) {
```
michax: Please add unit tests covering scenarios where supported body type is provided and it's not getting stringified.

ToreJuloe: Unit tests added.
I also had to update `package-lock.json` and change `testEnvironment` to `jsdom`.

PR #95, opened 2019-04-11
Improve versioning workflow

ToreJuloe:
Added some scripts to make versioning easier.

`prepare` will run after the module is installed and this should eliminate the need to have the final dist in the published npm module.

`preversion` _Run BEFORE bumping the package version_
Will make sure that all deps are installed and that linting and tests pass.

`version` _Run AFTER bumping the package version, but BEFORE commit_
Will stage `package-lock.json` to be committed, so the bumped version number is included.

`postversion` _Run AFTER bumping the package version, and AFTER commit_
Will push code and the new tag

Official documentation for the scripts [here](https://docs.npmjs.com/misc/scripts)

Now [publishing](https://docs.npmjs.com/updating-your-published-package-version-number) a new [version](https://docs.npmjs.com/cli/version.html) should be as easy as:
```
npm version patch -m "Upgrade to %s for reasons"
npm publish
```

Also added `npm-run-all` for Windows compatibility.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1432,6 +1432,24 @@
       "integrity": "sha1-jCpe8kcv2ep0KwTHenUJO6J1fJM=",
       "dev": true
     },
+    "array-filter": {
+      "version": "0.0.1",
+      "resolved": "https://registry.npmjs.org/array-filter/-/array-filter-0.0.1.tgz",
+      "integrity": "sha1-fajPLiZijtcygDWB/SH2fKzS7uw=",
+      "dev": true
+    },
+    "array-map": {
+      "version": "0.0.0",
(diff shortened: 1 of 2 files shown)


lukeapage:
Awesome! Can you add your script to the readme e.g.

## publishing

```
npm version patch -m "Upgrade to %s for reasons"
npm publish
```

michax:
Nice! 

As we have changelog, I would suggest keeping message as simple/default with just version. 

But maybe it's worth dropping changelog and instead use more descriptive commit messages on why version was bumped? 

@lukeapage @ToreJuloe Any suggestions here? Dropping update of chanelog has advantage of not 'pre/post' steps required. We could probably also automatize it if required by appending message to changelog.

PR #107, opened 2019-12-11
avoid mutating header options

mvjuul:
Some transports may append to the headers object. This can cause trouble since it will mutate the passed `options.headers` object.
This change creates a new `headers` object before passing it on, avoiding potential for mutation.

### DIFF
--- a/src/openapi/streaming/subscription.js
+++ b/src/openapi/streaming/subscription.js
@@ -73,7 +73,11 @@ function subscribe() {
         ReferenceId: referenceId,
         KnownSchemas: this.parser.getSchemaNames(),
     });
-    const options = { body: data, headers: extend({}, this.headers) };
+    const options = { body: data };
+
+    if (this.headers) {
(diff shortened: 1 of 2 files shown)


mvjuul:
Pending tests

src/openapi/streaming/subscription.js
```diff
@@ -73,19 +73,14 @@ function subscribe() {
         ReferenceId: referenceId,
         KnownSchemas: this.parser.getSchemaNames(),
     });
+    const options = { body: data, headers: Object.assign({}, this.headers) };
```
michax: Let's for now use extend method from object utils to make sure we are backward compatible with IE 10.
I'll work separately on either moving up min browser support or adding babel transpilation for ES 2015.

Extend is available as part of object utils:
image

And this is example usage for cloning subscription args:
image

mvjuul: Good call 👍

PR #110, opened 2020-03-05
Add more details to selected sub reset error logs

michax:
Add more details for error logs related to subscription reset and subscription errors.
This will make it easier to investigate live occurrences for this errors.

### DIFF
--- a/src/openapi/streaming/subscription.js
+++ b/src/openapi/streaming/subscription.js
@@ -349,7 +349,11 @@ function onUnsubscribeSuccess(referenceId, response) {
  */
 function onUnsubscribeError(referenceId, response) {
     if (referenceId !== this.referenceId) {
-        log.error(LOG_AREA, 'Received an error response for unsubscribing a subscription that has afterwards been reset - ignoring');
+        log.error(
+            LOG_AREA,
+            'Received an error response for unsubscribing a subscription that has afterwards been reset - ignoring',
+            { response, url: this.url, newReferenceId: referenceId, oldReferenceId: this.referenceId }
+        );
         return;
     }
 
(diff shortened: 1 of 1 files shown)


src/openapi/streaming/subscription.js
```diff
@@ -349,7 +349,11 @@ function onUnsubscribeSuccess(referenceId, response) {
  */
 function onUnsubscribeError(referenceId, response) {
     if (referenceId !== this.referenceId) {
-        log.error(LOG_AREA, 'Received an error response for unsubscribing a subscription that has afterwards been reset - ignoring');
+        log.error(
```
lukeapage: I'm not sure this should be an error.. 

also the message is incorrect - a unsubscribe,  subscribe could cause a change in reference id - the same as a reset.
when a reset occurs do we unsubscribe to be sure? because the subscription might have been cleaned up already

Maybe change it to a warning? but adding more info can't hurt.

src/openapi/streaming/subscription.js
```diff
@@ -379,7 +383,11 @@ function onModifyPatchSuccess(referenceId, response) {
  */
 function onModifyPatchError(referenceId, response) {
     if (referenceId !== this.referenceId) {
-        log.error(LOG_AREA, 'Received an error response for modify patch a subscription that has afterwards been reset - ignoring');
+        log.error(
```
lukeapage: same as above

src/openapi/streaming/subscription.js
michax: My initial intention was to have deeper view inside the log alerts we get for this scenarios and check why exactly they are happening. However, I might as well check it by manually reviewing kibana logs on WARN level instead of waiting for log alerts.

lukeapage:
I think this work is now obsolete?

PR #111, opened 2020-03-10
Allow formatting price up to 8 decimals

robatwilliams:
Adds a new format flag `UseExtendedDecimals` which will add up to 8 non-zero decimals beyond the requested number of decimals. The extra decimals will be part of either First/Pips or DeciPips depending on presence of `AllowDecimalPips`.

To support internal user story 1163324 and the PR I'm about to create for it, see there for context.

I've temporarily committed the `dist` folder to allow adding this branch as a Yarn dependency direct from GitHub, so probably don't need a beta version published.

@michax

### DIFF
--- a/src/price-formatting/format.js
+++ b/src/price-formatting/format.js
@@ -15,7 +15,8 @@ const NO_BREAK_SPACE = String.fromCharCode(0xa0);
 // -- Local methods section --
 
 function getFirstAndPipsParts(price, parts, numberFormatting) {
-    const minSize = (price.indexOf(numberFormatting.decimalSeparator) < 0 ? 2 : 3);
+    const minSize =
+        price.indexOf(numberFormatting.decimalSeparator) < 0 ? 2 : 3;
     if (price.length > minSize) {
         let pc = 2;
         parts.Pips = price.substr(price.length - pc, pc);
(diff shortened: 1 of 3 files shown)


dist/release/openapi-client.js
```diff
@@ -0,0 +1,6640 @@
+﻿/*
```
robatwilliams: Temporary, to remove

robatwilliams:
The internal PR that uses this is now ready for merge (it and this have been reviewed, and tested) once this is released.

src/price-formatting/format.js
```diff
@@ -155,31 +155,40 @@ function getFractionParts(formatFlags, basePart, deciPipsPart, numberFormatting)
...
     }
 
-    if (formatFlags.NoRounding) {
-        actualDecimals = numberFormatting.getActualDecimals(value);
-        if (actualDecimals <= decimals) {
-            formatFlags.NoRounding = false;
-        }
+    const actualDecimals = numberFormatting.getActualDecimals(value);
+
+    if (formatFlags.UseExtendedDecimals && getFlagPipDecimals(formatFlags) === 0) {
```
michax: Shouldn't this part be part of getPipDecimals method? This split is a little bit confusing. Would be good to have one clear getPipDecimals which would return required pipDecimals based on passed arguments.

robatwilliams: There are two cases - when `Allow[Two]DecimalPips` is specified, and when it's not. We want to add the extra decimals in a different location depending on this.

This here is to handle when we don't have `Allow[Two]DecimalPips`. We need to increase the decimals as much as we can up to 8, avoiding trailing zeroes (but keeping them up to specified decimals), so they end up in `First`. `Pips` always remains as two digits.

The case where we do have `Allow[Two]DecimalPips` is handled in `getPipDecimals()` and the extra digits will go into `DeciPips`.

PR #113, opened 2020-03-13
Add delay levels for streaming total reconnects.

michax:
Add delay levels for streaming total reconnects.
Allows client to provide various delay levels for specific retry counts levels. 

Example Structure and documented side effects on delays:
```
[
   { level: 0, delay: 1000 }, // First reconnect will wait 1s.
   { level: 1, delay: 2000 }, // Second reconnect will wait 2s.
   { level: 5, delay: 3000 }, // From second reconnect to 4th one, we will use 2s delay, and from 5th up to infinity, 3s delay would be used.
]
```

### DIFF
--- a/src/openapi/streaming/streaming.js
+++ b/src/openapi/streaming/streaming.js
@@ -91,7 +91,7 @@ function setNewContextId() {
  * @param defaultDelay {number} - The default delay.
  * @returns {number} Matching delay to retry index/try/count.
  */
-function findRetryDelay(retryLevels, retryIndex, defaultDelay) {
+export function findRetryDelay(retryLevels, retryIndex, defaultDelay) {
     let lastFoundDelay = defaultDelay;
 
     for (let i = 0; i < retryLevels.length; i++) {
@@ -152,7 +152,6 @@ function onConnectionStateChanged(nextState) {
             break;
 
         case this.CONNECTION_STATE_RECONNECTING:
-            this.retryCount = 0;
             updateConnectionQuery.call(this);
 
             this.orphanFinder.stop();
--- a/src/openapi/streaming/streaming.js
+++ b/src/openapi/streaming/streaming.js
@@ -84,10 +84,38 @@ function setNewContextId() {
 }
 
 /**
+ * Find matching delay based on current retry count/index.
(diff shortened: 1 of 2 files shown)


src/openapi/streaming/streaming.js
```diff
@@ -83,11 +83,39 @@ function setNewContextId() {
...
+/**
+ * Find matching delay based on current retry count/index.
+ * @param retryLevels - The retry levels that contain different delays for various retry count levels.
+ *                      Structure: [ { level: Number, delay: Number } ].
+ * @param retryIndex - The current retry index/try/count.
+ * @param defaultDelay {number} - The default delay.
+ * @returns {number} Matching delay to retry index/try/count.
+ */
+function findRetryDelay(retryLevels, retryIndex, defaultDelay) {
+    let lastFoundDelay = defaultDelay;
```
lukeapage: what if it gets past the levels - it shouldn't stay on the last item?

src/openapi/streaming/streaming.js
```diff
@@ -120,19 +148,18 @@ function onConnectionStateChanged(nextState) {
             for (let i = 0; i < this.subscriptions.length; i++) {
                 this.subscriptions[i].onConnectionUnavailable();
             }
-
             retryConnection.call(this);
             break;
 
         case this.CONNECTION_STATE_RECONNECTING:
-
+            this.retryCount = 0;
```
lukeapage: I thought it would only be necessary on conntected?

src/openapi/streaming/streaming.js
michax: It does. Current implementation uses last matching level and it's doesn't use exact matching. Instead it uses >=.

I've added full unit test coverage for all cases for both this method and streaming disconnect->retry lifecycle. 

This specif case is covered here:
https://github.com/SaxoBank/openapi-clientlib-js/pull/113/files#diff-30a8f287d3c66269a64c11675a8ce443R146

src/openapi/streaming/streaming.js
michax: Yeap, it was overkill. Dropped.

PR #115, opened 2020-03-14
fix for streaming 401

lukeapage:
If we get a disconnect from streaming while the token is valid but open api thinks it is invalid (because the users date is messed up or the time interval is not set), then we just keep trying to connect to streaming forever, ignoring that we get a 401. Because we never connect we never send a rest request that would kick an early auth token request.

I've had a go at initial changes to implement this.

Needed:

- [x] Unit tests
- [x] Testing
- [x] How to do this in raw websockets - do we get a websocket error? what tells us its unauthorized?
- [x] This is a breaking change, up the major and add a changelog - now we expect streaming to be given transportAuth instead of auth.
- [x] check coverage
- [x] pull out eslint change to reduce noise


src/openapi/transport/auth.js
```diff
@@ -6,150 +6,31 @@
...
-
 function makeTransportMethod(method) {
     return function(serviceGroup, urlTemplate, templateArgs, options) {
-        options = options || {};
-        options.headers = options.headers || {};
-        options.headers.Authorization = this.auth.getToken();
-
-        return this.transport[method](serviceGroup, urlTemplate, templateArgs, options)
-            .catch(onTransportError.bind(this, this.auth.getExpiry()));
+        const newOptions = {
```
lukeapage: @michax just a guess - but I think this might be why we are getting the batch errors with different auth token - because a token is retried and ends up in a batch but this has mutated the headers

src/openapi/authProvider.js
```diff
@@ -0,0 +1,302 @@
+/**
```
lukeapage: This was pulled out of transportAuth

Its basically the Auth class from inside there, but with token management and transport related concerns left in transportAuth

lukeapage: (apart from this there are no functional changes in this file)

src/openapi/batch-util.spec.js
```diff
@@ -300,7 +300,7 @@ describe('batch building', () => {
                     '--+--',
                     ''),
                 boundary: '+',
-            }
+            },
```
lukeapage: eslint did this when I set the ecma version to 2019

src/openapi/streaming/connection/connection.js
```diff
@@ -39,11 +39,12 @@ function onTransportFail(error) {
 
     this.transport.setReceivedCallback(this.receiveCallback);
     this.transport.setStateChangedCallback(this.stateChangedCallback);
+    this.transport.setUnauthorizedCallback(this.unauthorizedCallback);
     this.transport.setConnectionSlowCallback(this.connectionSlowCallback);
     this.transport.setErrorCallback(this.errorCallback);
 
     if (this.state === STATE_STARTED) {
-        this.transport.updateQuery(this.authToken, this.contextId);
+        this.transport.updateQuery(this.authToken, this.contextId, this.authExpiry);
```
lukeapage: we pass the authExpiry into the transport so it can pass it back when there is an error to tell us which token was used that generated a auth error.
The trouble is that signal-r would require a few internal changes to tell us this, so signal-r doesn't use it. I assume I can make raw websockets work withit.

src/openapi/streaming/connection/transport/signalr-transport.js
```diff
@@ -37,6 +37,9 @@ function mapConnectionState(state) {
  */
 function handleError(errorDetail) {
     log.error(LOG_AREA, 'Transport error', errorDetail);
+    if (errorDetail && errorDetail.source && errorDetail.source.status === 401) {
```
lukeapage: I looked at the signal-r source and also our logs to work out this condition 🤞

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -312,9 +318,12 @@ WebsocketTransport.prototype.getAuthorizePromise = function(contextId) {
                 resolve(response);
             })
             .catch((error) => {
+                if (error && error.status === 401) {
```
lukeapage: remove this - because the rest transport will already report the 401

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -170,6 +170,14 @@ function handleSocketClose(event) {
             code: event.code,
         });
 
+        // 1006 doesn't tell us its a 401 unauthorized, just that it might be
+        // the websocket spec disallows reading status codes for security reasons
+        // (so websocket can't be used to probe http endpoints)
+        // So, before we reconnect, lets re-authorize to be sure
+        if (event.code === 1006) {
+            this.getAuthorizePromise(this.contextId, true);
```
lukeapage: the idea being that because authorize uses the auth transport, a 401 will be handled through the auth transport

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -296,30 +308,31 @@ WebsocketTransport.prototype.setConnectionSlowCallback = function(callback) {
     this.connectionSlowCallback = callback;
 };
 
-WebsocketTransport.prototype.getAuthorizePromise = function(contextId) {
-    if (this.isAuthorized) {
+WebsocketTransport.prototype.getAuthorizePromise = function(contextId, forceAuthenticate) {
+    if (!forceAuthenticate && this.authorizePromise) {
```
lukeapage: I changed this logic - if we get a d/c around a token change we can be running 2 authorizations with the same parameters.
I can't think of how it would go wrong, but I don't like it..

src/openapi/streaming/streaming.js
```diff
@@ -35,35 +35,48 @@ const ignoreSubscriptions = {};
...
+    this.authProvider.tokenRejected(expiryOfRejectedToken);
 }
 
 /**
  * Reconnects the streaming socket when it is disconnected
  */
-function reconnect() {
-    if (this.connectionState !== this.CONNECTION_STATE_DISCONNECTED) {
-        log.error(LOG_AREA, 'Only call reconnect on a disconnected streaming connection');
+function connect() {
```
lukeapage: I combined connect and reconnect and made it wait for auth before trying to connect

.eslintrc
```diff
@@ -1,7 +1,7 @@
 {
     "root": true,
     "parserOptions": {
-        "ecmaVersion": 6,
+        "ecmaVersion": 2019,
```
jumoel: Is this change necessary to fix 401's?

src/openapi/streaming/connection/transport/websocket-transport.js
jumoel: Should this comment also be part of the code?

src/openapi/transport/queue.js
```diff
@@ -173,17 +172,23 @@ TransportQueue.prototype.emptyQueue = function() {
...
+            item.resolve(...args);
         },
-        function(result) {
-            if (this.transportAuth && result && result.status === 401) {
-                this.isQueueing = true;
+        (result, ...args) => {
+            if (this.authProvider && result && result.status === 401) {
                 this.addToQueue(item);
+                // if we are fetching a new token, wait
+                if (this.authProvider.isFetchingNewToken()) {
```
lukeapage: This is another change - I suspect that if we get a 401 but we already have a new token, we wouldn't be fetchng and the rest call will be hanging until the next token comes in - that could be disastrous....

.eslintrc
lukeapage: It has been pulled out to a seperate pr, merged to master and rebased on.

Its because I wanted to use object spread...

src/openapi/streaming/connection/transport/websocket-transport.js
lukeapage: yes you're right, added.

src/openapi/transport/auth.js
michax: Yeah, would match scenarios from live where we have both normal 401 failures followed by ones causes by subrequest containing auth token header.

src/openapi/transport/retry.spec.js
```diff
@@ -55,15 +55,15 @@ describe('openapi TransportRetry', () => {
         });
         setTimeout(() => {
             expect(transport.delete.mock.calls.length).toEqual(1);
-            transport.deleteReject(new TypeError('First Network request failed'));
+            transport.deleteReject({ status: 503, response: 'first' });
```
michax: Could you explain why we changed this test to status: 503 from TypeError?

src/openapi/streaming/streaming.js
```diff
@@ -35,35 +35,48 @@ const ignoreSubscriptions = {};
...
+        setNewContextId.call(this);
+        updateConnectionQuery.call(this);
 
-    this.reconnecting = true;
-    this.connection.start(onConnectionStarted.bind(this));
+        this.connection.start(onConnectionStarted.bind(this));
+    };
+
+    const expiry = this.authProvider.getExpiry();
+    if (expiry && expiry < Date.now()) {
```
michax: In which scenarios expiry might be missing? Do we want to connect in such scenarios?

src/openapi/transport/auth.js
```diff
@@ -227,77 +70,30 @@ function onTransportError(oldTokenExpiry, result) {
...
-    this.retryDelayMs = options && options.retryDelayMs || DEFAULT_RETRY_DELAY_MS;
-    this.maxRetryCount = options && options.maxRetryCount || DEFAULT_MAX_RETRY_COUNT;
-    this.maxAuthErrors = options && options.maxAuthErrors || DEFAULT_MAX_AUTH_ERRORS;
+function TransportAuth(baseUrl, authProvider, options) {
+    if (!authProvider) {
+        throw new Error('transport auth created without a auth provider');
+    }
+
     this.authErrorsCleanupDebounce = options && options.authErrorsCleanupDebounce || DEFAULT_AUTH_ERRORS_CLEANUP_DEBOUNCE;
+    this.maxAuthErrors = options && options.maxAuthErrors || DEFAULT_MAX_AUTH_ERRORS;
```
michax: I wonder if we should increase default value of DEFAULT_MAX_AUTH_ERRORS. Currently it's only 3 retries which happen one after another. On the other hand, not sure if we want to ddos our servers/balancers with retries.

src/openapi/transport/retry.spec.js
lukeapage: These type errors don't have a status or a isNetworkError - so they aren;t things we should really be retrying..
I've changed the logic so we only retry if there is a status and we match it.
Previously if the error did not have a status it would always retry.. thats highly dangerous.

src/openapi/streaming/connection/transport/signalr-transport.js
lukeapage: I only see 401. this one depends on what open api returns, which doesn't seem to ever be 403.
really the token refresh should probably be 401 only too.

src/openapi/streaming/streaming.js
lukeapage: Not for saxotrader, but I think we allow other users of the library to just pass in a token.

lukeapage: It seems we require an expiry, I will rmeove the if

src/openapi/transport/auth.js
lukeapage: We override it to Infinite and have done a long time. Not sure about the library default

src/openapi/streaming/streaming.js
lukeapage: *remove the `expiry &&` part

src/openapi/transport/queue.js
lukeapage: we listen to a token received event and when we get it received, the queue is emptied.. so if a token refresh fails nothing happens.

PR #121, opened 2020-03-19
use prettier

lukeapage:


.prettierignore
```diff
@@ -0,0 +1,4 @@
+node_modules
+dist
+package-lock.json
+libs
```
michax: Maybe worth duplicating excludes from .gitignore?
```
.grunt
node_modules
tmp
bin
obj
dist
.idea
.eslintcache
```

lukeapage: I tidied up the `gitignore` as many of these aren't valid any more and added the last 2 to prettierignore.

PR #123, opened 2020-03-19
Better unit test covering scenario where reconnect happens without prev. connect

michax:
Better unit test covering scenario where reconnect happens without prev. connect.


lukeapage:
@michax is this worth keeping or do you want to abandon it?

PR #159, opened 2020-04-20
Fix websocket transport

yugandhar02:
Fixes:

- null contextId while reconnecting.

- Fixes reconnection logic. i.e. don't fire disconnect event unless reconnect timeout is over

- Handle 1002 on close. i.e. re-authorise if token expired

### DIFF
--- a/src/openapi/streaming/connection/transport/websocket-transport.js
+++ b/src/openapi/streaming/connection/transport/websocket-transport.js
@@ -31,7 +31,7 @@ function normalizeWebSocketUrl(url) {
     return url.replace('http://', 'ws://').replace('https://', 'wss://');
 }
 
-function connect() {
+function createSocket() {
     try {
         const url = normalizeWebSocketUrl.call(
             this,
@@ -54,7 +54,7 @@ function connect() {
     }
 }
 
-function disconnect() {
+function destroySocket() {
     const socket = this.socket;
     if (!socket) {
         return;
@@ -80,8 +80,8 @@ function reconnect() {
     this.reconnectTimeout = setTimeout(() => {
         this.reconnectCount++;
 
-        disconnect.call(this);
-        connect.call(this);
+        destroySocket.call(this);
+        createSocket.call(this);
 
         log.debug(LOG_AREA, 'Transport reconnected');
     }, DEFAULT_RECONNECT_DELAY);
@@ -153,14 +153,14 @@ function parseMessage(rawData) {
  * @param { Object } error - The error object with message property.
  */
 function handleFailure(error) {
-    disconnect.call(this);
+    destroySocket.call(this);
     this.stateChangedCallback(constants.CONNECTION_STATE_FAILED);
     this.failCallback(error);
 }
 
 function handleSocketOpen() {
     if (this.socket) {
-        this.hasWorked = true;
+        this.hasBeenConnected = true;
         this.reconnectCount = 0;
 
         log.debug(LOG_AREA, 'Socket opened');
@@ -193,7 +193,7 @@ function handleSocketClose(event) {
         return;
     }
 
-    if (!this.hasWorked) {
+    if (!this.hasBeenConnected) {
         handleFailure.call(this, {
             message: `websocket error occured. code: ${event.code}, reason: ${event.reason}`,
         });
@@ -203,7 +203,7 @@ function handleSocketClose(event) {
 
     const isCleanDisconnect = event.wasClean === true;
     if (!isCleanDisconnect) {
-        log.error(LOG_AREA, 'websocket connection closed abruptly', {
+        log.warn(LOG_AREA, 'websocket connection closed abruptly', {
             readyState: this.socket.readyState,
             code: event.code,
             reason: event.reason,
@@ -248,7 +248,7 @@ function WebsocketTransport(baseUrl, restTransport, failCallback = NOOP) {
     this.authorizeServiceGroup = 'streamingws';
 
     // If true, indicates that transport had at least once successful connection (received onopen).
-    this.hasWorked = false;
+    this.hasBeenConnected = false;
 
     this.lastMessageId = null;
     this.reconnectTimeout = null;
@@ -352,7 +352,7 @@ WebsocketTransport.prototype.start = function(options, callback) {
     }
 
     if (this.socket) {
-        log.debug(LOG_AREA, 'only one socket per connection is allowed');
+        log.warn(LOG_AREA, 'only one socket per connection is allowed');
         return;
     }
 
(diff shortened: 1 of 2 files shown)


src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -13,53 +13,64 @@ const LOG_AREA = 'PlainWebSocketsTransport';
...
+    this.socket = null;
 }
 
 function reconnect() {
     if (this.reconnectCount >= DEFAULT_RECONNECT_LIMIT) {
-        handleFailure.call(this, {
-            message:
-                'Transport reached maximum amount of reconnects allowed till last successful connection.',
-        });
+        this.stop();
```
yugandhar02, line 73: This will enable infinite retry for plain websockets
should this be controlled by flag ?
whether to fallback to signalR or retry again

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -181,58 +189,34 @@ function handleSocketMessage(messageEvent) {
...
+    if (!this.socket) {
+        return;
     }
-}
 
-function handleSocketOpen() {
-    if (this.socket && this.socket.readyState === STATE_INTERNAL_OPEN) {
-        this.hasWorked = true;
-        this.reconnectCount = 0;
+    if (!this.hasWorked) {
```
lukeapage, line 196: hasWorked - maybe rename to hasBeenConnected ? or whatever it means - something more specific

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -181,58 +189,34 @@ function handleSocketMessage(messageEvent) {
...
-}
 
-function handleSocketError(error) {
-    if (this.socket) {
-        log.error(LOG_AREA, 'Socket error occurred', {
-            code: error.code,
-            reason: error.reason,
+    const isCleanDisconnect = event.wasClean === true;
+    if (!isCleanDisconnect) {
+        log.error(LOG_AREA, 'websocket connection closed abruptly', {
```
lukeapage, line 206: warning?
Its not an error from our perspective to lose a internet connection

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -319,15 +303,6 @@ WebsocketTransport.prototype.setErrorCallback = function(callback) {
     this.errorCallback = callback;
 };
 
-/**
- *
- * Suggested next step is a fallback to another transport if possible.
- * @param callback
- */
-WebsocketTransport.prototype.fail = function(callback) {
```
lukeapage, line 327: Was this unused?

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -370,37 +345,40 @@ WebsocketTransport.prototype.start = function(options, callback) {
...
-        this.destroy();
-        log.debug(LOG_AREA, 'Transport stopped');
-    }
+    disconnect.call(this);
+
+    clearInterval(this.reconnectTimeout);
+    this.reconnectTimeout = null;
+    this.contextId = null;
+    this.lastMessageId = null;
+    this.authorizePromis = null;
```
lukeapage, line 377: typo - missing e

src/openapi/streaming/connection/transport/websocket-transport.js
yugandhar02: yup

src/openapi/streaming/connection/transport/websocket-transport.js
yugandhar02: sure. will update

src/openapi/streaming/connection/transport/websocket-transport.js
michax: Yes, we only assign it in constructor and we don't use this method, so fine to remove it.

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -181,58 +189,34 @@ function handleSocketMessage(messageEvent) {
...
-        // and if that gets a 401, that will report the token invalid via
-        // authTransport
-        if (event.code === 1006) {
-            this.getAuthorizePromise(this.contextId, true);
-        }
-
-        // Order here is important. We need to invoke state callback first, as parent streaming manager updates query upon disconnect.
-        this.stateChangedCallback(constants.CONNECTION_STATE_DISCONNECTED);
-        reconnect.call(this);
+    if (!this.socket) {
```
michax, line 192: In which scenario is it possible to get empty this.socket state? I don't think it's possible or should be. If we plan to clean the this.socket instance, we should also remove callbacks before hand.

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -181,58 +189,34 @@ function handleSocketMessage(messageEvent) {
...
-        } else {
-            handleFailure.call(this, {
-                message: `WebSocket connection failed with error code: ${error.code}, reason: ${error.reason}`,
-            });
-        }
+    if (event.code === socketCloseCodes.TOKEN_EXPIRED) {
+        this.unauthorizedCallback();
     }
+
+    reconnect.call(this);
```
michax, line 219: Reconnecting immediately after calling this.unauthorizedCallback doesn't look like a good approach. If current token is invalid, new connect will still fail and might cause spam of oauth token re-requests.

I think instead we should postpone reconnect here until fresh token is received.

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -370,37 +345,40 @@ WebsocketTransport.prototype.start = function(options, callback) {
...
-        log.debug(LOG_AREA, 'Creating WebSocket connection', { url });
-        this.socket = new WebSocket(url);
-        configure.call(this);
         this.startedCallback();
+        this.stateChangedCallback(constants.CONNECTION_STATE_CONNECTING);
+        connect.call(this);
     });
 };
 
 WebsocketTransport.prototype.stop = function() {
```
michax, line 370: Nitpick. My suggestion would be to current stop rename to disconnect and existing disconnect to rename to destroy or close. Reading the code I assumed all disconnect method invocations also caused state changed to disconnected.

src/openapi/streaming/connection/transport/websocket-transport.js
michax: Is it possible to happen in current disconnect method where we are cleanly disconnecting?

yugandhar02: nope... unless we call this method by mistake
it has reconnection logic and it could lead to hair pulling bugs
Earlier it was guarded too and I thought it was for same reason ?

src/openapi/streaming/connection/transport/websocket-transport.js
yugandhar02: reconnection happens after reconnect delay
we were relying on timer for authorisation before this change as well
I have already discussed this on slack with you

src/openapi/streaming/connection/transport/websocket-transport.js
yugandhar02: I am using same interface as in signalR transport.
the main confusion here is due to the fact that we are changing streaming state instead of transport
It has 1:1 maping with singalR

src/openapi/streaming/connection/transport/websocket-transport.js
yugandhar02: @lukeapage Can you please check this comment ?

src/openapi/streaming/connection/transport/websocket-transport.js
michax: I don't think previous approach was right :) Can we try to fix it by making reconnect logic wait for authorization response? When we receive new token, we currently call: WebsocketTransport.prototype.updateQuery.
This would be indicator that we received new token after token experied close code and we are ready to continue re-connection. 

WDYT?

yugandhar02: Done

PR #204, opened 2020-06-11
Fix stackoverflow exception during JSON payload parsing

yugandhar02:


src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -25,8 +25,50 @@ const DEFAULT_RECONNECT_LIMIT = 10;
...
+        recurse();
+    } catch (e) {
+        maxCallstackSize = index;
+    }
+
+    return maxCallstackSize;
+}
+
+function getJSONPayloadString(payloadBuffer) {
+    const chunkSize = getMaxCallstackSize();
```
lukeapage, line 53: I don't think we should attempt to get a maximum call stack - its an expensive operation and dangerous - I've seen older browsers crash and sometimes be unable to catch the exception.

I compared some techniques and even setting the chunk size to 150 its not that much difference from whatever max call size is. at 350 chunk size its 20% slower.. so a pick a number thats definitely very safe - for instance I'd be happy with 1000 ?

image

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -25,8 +25,50 @@ const DEFAULT_RECONNECT_LIMIT = 10;
...
+    const chunks = Math.ceil(payloadBuffer.length / chunkSize);
+    let payload = '';
+    let chunkIndex = 0;
+
+    while (chunkIndex < chunks) {
```
lukeapage, line 58: we should have a test covering this - which will also be easier with a smaller chunk size

src/openapi/streaming/connection/transport/websocket-transport.js
michax, line 53: If we decide to use staticly defined max call stack, please validate it against mobile. For example form me, payload of 1000 entires is crashing my chrome on my 6 gb anroid phone. 

image

Test run:
https://jsperf.com/animuj-fromcharcode

src/openapi/streaming/connection/transport/websocket-transport.js
yugandhar02: sure
will add it

src/openapi/streaming/connection/transport/websocket-transport.js
yugandhar02: https://jsperf.com/yugk-fromcharcode
1000 seems good enough

yugandhar02: https://jsperf.com/yugk-fromcharcode
I have tested with 4gb android phone
works fine with 30k entries and chunksize of 500, 1000, 1500

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -135,9 +157,17 @@ function parseMessage(rawData) {
             const payloadBuffer = new Uint8Array(
                 rawData.slice(index, index + payloadSize),
             );
-            data = String.fromCharCode.apply(null, payloadBuffer);
-            data = decodeURIComponent(escape(data));
```
yugandhar02, line 139: @lukeapage Do we really need to sanitise the data
escape is deprecated now
I have discussed it a bit with @michax and removed

yugandhar02: As discussed on slack. Can be removed

PR #212, opened 2020-06-25
Fix typed array slice method not supported exception in IE11 and old chrome

yugandhar02:


src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -31,16 +31,18 @@ function getJSONPayloadString(payloadBuffer) {
...
     let payload = '';
     let chunkIndex = 0;
 
     while (chunkIndex < chunks) {
         payload += String.fromCharCode.apply(
             null,
-            payloadBuffer.slice(
-                chunkIndex * chunkSize,
-                (chunkIndex + 1) * chunkSize,
+            new Uint8Array(
```
lukeapage, line 41: whats the significance of when to wrap the result of slice in a Uint8Array?

PR #254, opened 2020-09-17
Revew log messages and improve network resiliency

lukeapage:
I've reviewed all logs - Searching for occurrences of our errors/warns and evaluating whether they should be a error/warning on the basis that error/warns should be indications that something is gone wrong that is abnormal. If a network error occurs or a fairly normal event that isn't a problem occurs, including the effect of a single open api server going down and subscriptions failing over then it should not be a warn/error.

### DIFF
--- a/src/openapi/streaming/connection/transport/websocket-transport.js
+++ b/src/openapi/streaming/connection/transport/websocket-transport.js
@@ -69,31 +69,38 @@ function destroySocket() {
     this.socket = null;
 }
 
+function restartConnection() {
+    this.reconnectTimeout = null;
+    this.reconnectCount++;
+
+    createSocket.call(this);
+
+    log.debug(LOG_AREA, 'Transport reconnected');
+}
+
 function reconnect(isImmediate) {
     if (this.reconnectCount >= DEFAULT_RECONNECT_LIMIT) {
         this.stop();
         return;
     }
 
+    if (this.reconnectTimeout && !isImmediate) {
+        log.warn(LOG_AREA, 'Reconnecting when already reconnecting');
+    }
+
     clearTimeout(this.reconnectTimeout);
+    this.reconnectTimeout = null;
     this.stateChangedCallback(constants.CONNECTION_STATE_RECONNECTING);
 
-    const restartConnection = () => {
-        this.reconnectCount++;
-
-        destroySocket.call(this);
-        createSocket.call(this);
-
-        log.debug(LOG_AREA, 'Transport reconnected');
-    };
+    destroySocket.call(this);
 
     if (isImmediate) {
-        restartConnection();
+        restartConnection.call(this);
         return;
     }
 
     this.reconnectTimeout = setTimeout(
-        restartConnection,
+        restartConnection.bind(this),
         DEFAULT_RECONNECT_DELAY,
     );
 }
@@ -186,6 +193,8 @@ function handleSocketOpen() {
 }
 
 function handleSocketMessage(messageEvent) {
+    this.lastMessageTime = Date.now();
+
     if (messageEvent.data instanceof ArrayBuffer) {
         let parsedMessages;
         try {
@@ -211,6 +220,14 @@ function handleSocketMessage(messageEvent) {
         });
 
         this.receivedCallback(parsedMessages);
+    } else {
+        log.error(
+            LOG_AREA,
+            'Received a non-ArrayBuffer message',
+            {
+                payload: messageEvent.data,
+            },
+        );
     }
 }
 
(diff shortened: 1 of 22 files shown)


src/openapi/transport/auth.js
```diff
@@ -185,12 +192,19 @@ TransportAuth.prototype.incrementErrorCounter = function(url) {
  * @param {string} url - The url/endpoint for which error count is returned.
```
yugandhar02, line 192: update or remove comment ?

src/openapi/transport/auth.js
```diff
@@ -185,12 +192,19 @@ TransportAuth.prototype.incrementErrorCounter = function(url) {
...
-        return this.authorizationErrorCount[url];
+TransportAuth.prototype.areUrlAuthErrorsProblematic = function(
+    url,
+    authExpiry,
+) {
+    if (this.authorizationErrors.hasOwnProperty(url)) {
+        for (let i = 0; i < this.authorizationErrors[url].length; i++) {
+            if (this.authorizationErrors[url][i].authExpiry !== authExpiry) {
+                return true;
+            }
```
yugandhar02, line 203: so we will log error if request with same url with different token failed ?
may be we should update error message as well ?

src/openapi/transport/auth.js
```diff
@@ -155,28 +149,41 @@ TransportAuth.prototype.head = makeTransportMethod('head');
...
 
 /**
- * Run debounced cleanup of error counter map
+ * Cleanup of error counter map
  */
-TransportAuth.prototype.debounceErrorCounterCleanup = function() {
-    if (this.errorCleanupTimoutId) {
-        clearTimeout(this.errorCleanupTimoutId);
+TransportAuth.prototype.cleanupAuthErrors = function() {
+    const cleanThoseBefore = Date.now() - this.authErrorsIgnoreDuration;
```
yugandhar02, line 155: the name is confusing 
we are not actually ignoring the auth errors but keeping them for certain duration to check whether the req with same url with diff authtoken failed ?

src/openapi/transport/auth.js
lukeapage: yes. updated message. its to stop one problematic service making us ddos the token request url.

src/openapi/transport/auth.js
lukeapage: Tried to improve the names..

yugandhar02 (approved):

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -192,12 +191,16 @@ function handleSocketMessage(messageEvent) {
...
         try {
             parsedMessages = parseMessage.call(this, messageEvent.data);
-        } catch (e) {
-            handleFailure.call(this, {
-                message: `Error occurred during parsing of plain WebSocket message. Message: ${e.message}`,
-                payload: e.payload,
-                payloadSize: e.payloadSize,
-            });
+        } catch (error) {
+            log.error(
```
yugandhar02, line 195: log area is missing

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -360,29 +360,29 @@ WebsocketTransport.prototype.getAuthorizePromise = function(
...
-        const options = {
-            headers: {
-                'X-Request-Id': getRequestId(),
-                Authorization: authToken,
-            },
-        };
-        const url = `${this.authorizeUrl}?contextId=${contextId}`;
+    const options = {
+        headers: {
+            'X-Request-Id': getRequestId(),
```
yugandhar02, line 403: Do we only use this header for batching ?
or it also serve the purpose of tracking request ?
request id sent by openapi is different than this one ?

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -69,31 +69,38 @@ function destroySocket() {
...
+}
+
 function reconnect(isImmediate) {
     if (this.reconnectCount >= DEFAULT_RECONNECT_LIMIT) {
         this.stop();
         return;
     }
 
+    if (this.reconnectTimeout && !isImmediate) {
+        log.warn(LOG_AREA, 'Reconnecting when already reconnecting');
```
yugandhar02, line 88: return early ?

src/openapi/streaming/connection/transport/websocket-transport.js
```diff
@@ -252,6 +269,23 @@ function handleSocketClose(event) {
...
+    // if we haven't got a message recently
+    // but we found a orphan recently
+    // and got a network error subscribing
+    // and our reconnectCount is 0 so we are not currently reconnecting
+    if (this.lastMessageTime < fiveSecondsAgo &&
+        this.lastOrphanFound > fiveSecondsAgo &&
+        this.lastSubscribeNetworkError > fiveSecondsAgo &&
+        this.reconnectCount === 0) {
+
+        // reconnect immediately as no need to wait - this is the initial event
```
yugandhar02, line 284: can we have debug log here that we are reconnecting due to possible network error ?

src/openapi/streaming/connection/transport/websocket-transport.js
lukeapage: its just for tracking whether a request is a duplicate. I haven't changed this.

src/openapi/streaming/connection/transport/websocket-transport.js
lukeapage: I was a bit scared, but yes I think we should.

PR #264, opened 2020-10-09
Improve imports and fix tests

yugandhar02:
1. Improves imports. Uses absolute import. I have tested tree shaking.

1. Fixes / remove duplicate tests, expect statements

1. Add more linting rules for tests and import order

1. Fixes failing test due to max callstack size exceeded exception by fixing Math.random mock.
It was getting used for quick sort in source map. This eventually fixes failing travis ci build


src/openapi/streaming/streaming.spec.js
```diff
@@ -71,6 +66,7 @@ describe('openapi Streaming', () => {
         fetchMock = mockFetch();
 
         installClock();
+        mockMathRandom();
```
yugandhar02, line 69: this fixes max callstack exceeded error on local
the reason was Math.random is being used to get pivot index for quick sort in source map node modules
and since it always returns 0, quicksort runs with worst complexity

lukeapage (approved):
Looks much better! thanks.

PR #268, opened 2020-10-14
Use websocket for token renewal in signalr core

yugandhar02:


src/openapi/streaming/connection/transport/signalr-core-transport.js
```diff
@@ -191,27 +199,15 @@ SignalrCoreTransport.prototype.start = function(options, onStartCallback) {
...
-            }
+            log.error(
+                LOG_AREA,
+                'Error occurred while connecting to streaming service',
+                {
+                    error,
+                },
+            );
+
+            this.transportFailCallback();
```
lukeapage, line 210: so.. before it retried and now it fails so that it goes back to the on premise streaming?

I think it could be like this in the beginning, but eventually we are going to probably want it to retry some times and then fallback.

src/openapi/streaming/connection/transport/signalr-core-transport.js
```diff
@@ -334,58 +330,84 @@ SignalrCoreTransport.prototype.updateQuery = function(
...
-                return this.renewSession(authToken, contextId);
-            }
-
-            if (error.status === 401) {
-                this.unauthorizedCallback();
-                return;
-            }
-
-            log.error(
+            log.info(
```
lukeapage, line 401: should this be a warn or error? it is an unexpected scenario?

lukeapage (changes requested):

src/openapi/streaming/connection/transport/signalr-core-transport.js
yugandhar02: Yes.
Moreover I am looking into using fetch util as httpClient for http requests made by signalr lib.

src/openapi/streaming/connection/transport/signalr-core-transport.js
yugandhar02: this could occur when signalr lib is unable to send websocket message
it could be due to network error
I don't see it as unexpected scenario 
other scenarios could be streaming service is not available and in that case connection will be dropped

yugk: I will change it to warning

yugandhar02: will change it to warning

src/openapi/streaming/connection/transport/signalr-core-transport.js
lukeapage: what http requests are used by signalr app? what does this have to do with the above?

yugandhar02: negotiation request 
and signalr use it for long polling for fallback scenario.
our fetch util has isNetworkError as clear api
but I suppose we will be dealing with same problem ?

yugandhar02: That doesn't change things
So I will keep it as it is

src/openapi/streaming/connection/transport/signalr-core-transport.js
```diff
@@ -334,58 +330,84 @@ SignalrCoreTransport.prototype.updateQuery = function(
...
 };
 
 SignalrCoreTransport.prototype.renewSession = function(authToken, contextId) {
-    const options = {
-        headers: {
-            Authorization: authToken,
-        },
-    };
-    const url = `${this.baseUrl}/streaming/renewal/renewsession`;
+    if (!this.connection || this.isReconnecting) {
```
lukeapage, line 341: if a token expires whilst we are connecting or reconnecting, don't we need to remember to renew the session once we are connected?

src/openapi/streaming/connection/transport/signalr-core-transport.js
```diff
@@ -334,58 +330,84 @@ SignalrCoreTransport.prototype.updateQuery = function(
...
+                    log.warn(
+                        LOG_AREA,
+                        'Session not found while renewing session',
+                        {
+                            contextId,
+                        },
+                    );
+
+                    // should try to reconnect with different context id
+                    this.stop();
```
lukeapage, line 378: should we stop? what if then then gets called back but we are already reconnecting or have reconnected?

yugandhar02: I don't think that's even possible since we are using websocket messages.
if we are in reconnecting state that means this callback will get cancelled before triggering  reconnecting state
and for 2nd case where we are trying to renew when we are reconnecting, thats already being handled

src/openapi/streaming/connection/transport/signalr-core-transport.js
yugandhar02: we will reconnect with new token so I don't think thats needed 
I will still double check it

yugandhar02: there was issue with token factory
now fixed
so while reconnecting it will use new token

lukeapage: is this because the token is sent right at the end of connecting?

yugandhar02: Not sure what you mean by that
It uses token factory i.e. a normal function which returns updated token
lets consider scenario of reconnecting state
signalr is trying to reconnect and we got new token and trying to renew session
we store this token as instance variable. The same token will be served from token factory. so next time signalr tries to reconnect it will use this token factory to get updated token
since we reconnected with new token we don't need to renew it again.

lukeapage (approved):
Looks good!

PR #270, opened 2020-10-15
Support cloud services

robatwilliams:
Capability to configure particular services as cloud services. This sends them to `/oapi` rather than `/openapi`, and excepts them from HTTP-level batching.

```javascript
new SomethingTransport('https://www.saxotrader.com', {
  services: {
    'fooservice': { useCloud: true }  
  }
});
```

Service groups don't exist in the cloud so it's no longer universal. It only remains as a concept in the batch transport for on-prem only. Elsewhere it's just a path string with no particular meaning, so it's renamed to service path to reflect it could be either a service group or the name of a cloud service (causes a lot of diff noise after the first few commits).

In future we may extend the config with a `useBatching` option to allow our "modern services" to be run on-prem where they probably won't have batching either.

In the possible future case of a new service having the same name as an existing service group, the version can be included in the service path configuration and passed to the client in the service path rather than in the 2nd parameter as is usually done.

Internal references:
* PR: 87223
* Wiki: "Use of /oapi by SaxoTrader"

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,4 @@
-### v.8.0.0
+### v8.0.0
 - BREAKING - Add support for cloud /oapi services. The baseUrl used when constructing transports should no longer include "/openapi"; this is now added by the library.
 
 ### v7.2.1
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,6 @@
+### v.8.0.0
+- BREAKING - Add support for cloud /oapi services. The baseUrl used when constructing transports should no longer include "/openapi"; this is now added by the library.
+
 ### v7.2.1
 
 - Fix some log messages not passing LOG_AREA and add more details to a parsing error
--- a/src/openapi/transport/batch.js
+++ b/src/openapi/transport/batch.js
@@ -181,6 +181,7 @@ function TransportBatch(transport, baseUrl, options) {
         basePath += '/';
     }
 
+    // Batching is a service group level facility, so isn't applicable/available for /oapi
     this.basePath = basePath + 'openapi/';
 
     if (options && options.host) {
--- a/src/openapi/transport/batch.js
+++ b/src/openapi/transport/batch.js
@@ -180,7 +180,7 @@ function TransportBatch(transport, baseUrl, options) {
         basePath += '/';
     }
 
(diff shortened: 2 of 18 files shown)


src/openapi/transport/batch.js
```diff
@@ -222,6 +224,16 @@ TransportBatch.prototype.addToQueue = function(item) {
     }
 };
 
+/**
+ * @private
+ * @param item
+ */
+TransportBatch.prototype.shouldQueue = function(item) {
+    const serviceOptions = this.services[item.servicePath] || {};
```
lukeapage, line 232: optional chaining instead of creating an empty object?
if this uses preset-env it might require a babel upgrade though :S

src/openapi/transport/batch.js
```diff
@@ -180,7 +181,7 @@ function TransportBatch(transport, baseUrl, options) {
         basePath += '/';
     }
 
-    this.basePath = basePath;
+    this.basePath = basePath + 'openapi/';
```
lukeapage, line 184: comment that this cannot be used on oapi/ ?

lukeapage (changes requested):

src/openapi/transport/batch.js
robatwilliams: Yes it would upgrading various things including babel and eslint

lukeapage:
Sorry, will add new serviceGroup reference here: https://github.com/SaxoBank/openapi-clientlib-js/pull/281

please rebase and fix before merging.

src/openapi/transport/batch.js
robatwilliams: I think that's best not mixed in with this PR

robatwilliams:
Ready for merge now please.

src/openapi/transport/batch.js
lukeapage: yes thats ok

lukeapage (approved):

changelog.md
```diff
@@ -1,3 +1,6 @@
+### v.8.0.0
```
lukeapage, line 1: v8.0.0 - I made the pattern consistent in my last pr

PR #303, opened 2020-12-16
Send last received message id while reconnecting

yugandhar02:
This PR modifies reconnection logic to send last received message id while reconnecting.
Last received message id is sent as query string while trying to reconnect again.
Server will return _reconnect control message if session is not found while trying to reconnect. In this case client will disconnect and try to reconnect with new context id.

### DIFF
--- a/src/openapi/streaming/connection/transport/signalr-core-transport.js
+++ b/src/openapi/streaming/connection/transport/signalr-core-transport.js
@@ -4,7 +4,7 @@ import * as constants from '../constants';
 
 const LOG_AREA = 'SignalrCoreTransport';
 const NOOP = () => {};
-const RECONNECT_DELAYS = [2000, 3000, 5000, 10000, 20000];
+const RECONNECT_DELAYS = [0, 2000, 3000, 5000, 10000, 20000];
 
 const renewStatus = {
     SUCCESS: 0,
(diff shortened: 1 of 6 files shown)


src/openapi/streaming/connection/transport/signalr-core-transport.js
```diff
@@ -24,8 +24,19 @@ function handleLog(level, message) {
...
+function buildConnection({
+    baseUrl,
+    contextId,
+    messageId,
+    accessTokenFactory,
+    protocol,
+}) {
+    let queryString = `contextId=${contextId}`;
+    if (messageId) {
+        queryString = `${queryString}&&messageId=${messageId}`;
```
michax, line 36: Typo with double "&&"?

src/openapi/streaming/connection/transport/signalr-core-transport.js
```diff
@@ -89,18 +101,110 @@ function parseMessage(message, utf8Decoder) {
...
+        .catch((err) => {
+            log.debug(LOG_AREA, 'Failed to reconnect', {
+                error: err,
+            });
+
+            this.handleConnectionClosure();
+        });
+}
+
+function retryConnection(retryIndex, prevError) {
```
michax, line 180: I'm not in love with idea of introducing manual reconnect on top of what's implemented by signalr :( 
Few questions:

* Shouldn't oapi server be aware about last sent message and their 'received' status? Why they cannot calculate last message id themself? 
* Isn't there some 'factory' method for query building for reconnecting already built it into SignalrCore?

michax (changes requested):

src/openapi/streaming/connection/transport/signalr-core-transport.js
yugandhar02: will fix

src/openapi/streaming/connection/transport/signalr-core-transport.js
yugandhar02: server won't know if client is disconnected due to network issue so it won't have last received message.
I don't see any factory method for query build in new signalr core

yugandhar02: Although I see there is setter for baseUrl which I missed somehow and it can be changed during reconnection.
It should do the trick.
will update

michax (approved):

PR #304, opened 2021-01-05
Bug fix rounding decimals formatters

bkozar:
**Desription**
For some numbers when calling format functions (ex: priceFormatting.format), the number was rounded down while it should be up. 
Example: 
priceFormatting.format(1.005, 2) is returning 1.00 while it should be 1.01.

The issue is caused by binary floating-point representation. 
Example: 
- 0.1 + 0.2 !== 0.3
- 1.005 * 100 = 100.49999999999999

More about it:
https://www.codemag.com/Article/1811041/JavaScript-Corner-Math-and-the-Pitfalls-of-Floating-Point-Numbers

**Solution**
Changing the rounding to use the numbers represented in exponential notation.
http://www.jacklmoore.com/notes/rounding-in-javascript/

Example:
1.005 * 100 is converted to 1.005e2 which gives us 100.5 in a non exponential notation.

**Test**
Added unit test for the above example priceFormatting.format(1.005, 2).
All existing tests are passing.

### DIFF
--- a/src/number-formatting/format.js
+++ b/src/number-formatting/format.js
@@ -99,7 +99,6 @@ function expandNumber(number, precision, options) {
 
 function roundNumber(number, decimals) {
     // Shift with exponential notation to avoid floating-point issues.
-    // See [MDN](https://mdn.io/round#Examples) for more details.
     let pair = `${number}e`.split('e');
     const value = Math.round(`${pair[0]}e${Number(pair[1]) + decimals}`);
     pair = `${value}e`.split('e');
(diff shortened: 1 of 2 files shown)


lukeapage:
Maybe we need a more drastic implementation 
It’s interesting to compare with the lodash round : https://github.com/lodash/lodash/blob/master/.internal/createRound.js

lukeapage:
I’d rather use that approach given it’s been stable and bug less in lodash for years.

bkozar:
> I’d rather use that approach given it’s been stable and bug less in lodash for years.

Ok, used that approach.

PR #308, opened 2021-01-14
Detect extended asset types header in batch subrequest, and append to…

majtsaxo:
… batch transport if found

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "openapi-clientlib",
-    "version": "8.0.3",
+    "version": "8.0.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "openapi-clientlib",
-    "version": "8.0.3",
+    "version": "8.0.4-ci-extended-asset-types-header-batchrequest",
```
lukeapage, line 3: just update this to be 8.0.4 and looks good

majtsaxo: Updated to 8.0.4

PR #321, opened 2021-02-04
Fix various minor issues

lukeapage:
From analyzing logs on live I've found some bugs - primarily the problem is that isNetworkError is false when it should be true for alot of batch requests.

When I found the issue, I realised that we should not be getting this error at all under normal operation, so analysing that I found that when calls are delayed it can trigger the authorisation error overload when the calls were dispatched many minutes apart. So I fixed that by debouncing calls based on the time from when the request starts.

### DIFF
--- a/src/openapi/transport/auth.js
+++ b/src/openapi/transport/auth.js
@@ -31,13 +31,19 @@ function makeTransportMethod(method) {
             urlTemplate,
             templateArgs,
             newOptions,
-        ).catch(onTransportError.bind(this, this.authProvider.getExpiry()));
+        ).catch(
+            onTransportError.bind(
+                this,
+                this.authProvider.getExpiry(),
+                Date.now(),
+            ),
+        );
     };
 }
 
-function onTransportError(oldTokenExpiry, result) {
+function onTransportError(oldTokenExpiry, timeRequested, result) {
     if (result && result.status === 401) {
-        this.addAuthError(result.url, oldTokenExpiry);
+        this.addAuthError(result.url, oldTokenExpiry, timeRequested);
         this.cleanupAuthErrors();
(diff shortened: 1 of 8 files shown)


src/openapi/transport/batch.js
```diff
@@ -36,9 +36,10 @@ function emptyQueueIntoServiceGroups() {
 
 function batchCallFailure(callList, batchResponse) {
     const isAuthFailure = batchResponse && batchResponse.status === 401;
+    const isNetworkError =
+        !batchResponse || batchResponse.isNetworkError || !batchResponse.status;
```
lukeapage, line 40: it is still a mystery as to why I would need to guard status here. pending any good ideas.

src/openapi/transport/auth.js
```diff
@@ -52,7 +58,12 @@ function onTransportError(oldTokenExpiry, result) {
                 'Too many authorization errors occurred for different tokens within a specified timeframe for a specific endpoint',
                 result.url,
             );
-            return;
+            throw {
+                message: 'Auth overload',
+                // obscure the real status code so that the queue does not keep retrying
+                status: 500,
```
michax, line 64: For people analyzing logs 500 might be confusing. Maybe we can at least indicate in message that this is 'synthetic' error we create on our end in relation to failed authorization? I assume status needs to be the same as we have specific logic to handle various error codes and some of them can force re-tries?

michax (approved):
Left some small suggestion.

src/openapi/transport/batch.js
michax: @lukeapage  Do we still need this one with auth fix?

src/openapi/transport/auth.js
lukeapage: yep, its not ideal. fixed it.

PR #323, opened 2021-02-05
Handle reset better

lukeapage:
1. When we get a reset, unsubscribe and wait for that subscribe to finish before subscribing again
2. As a side effect - resets occurrring whilst unsubscribing have no effect
3. Because we could be in the process of subscribing so unsure if the subscription is reset or not.. we need to make sure we force a unsubscribe - so added a force option to unsubscribe
4. The force option on unsubscribe is the same as the existing modify_subscribe logic - its achieving the same thing by making sure a unsubscribe/subscribe happens. So remove that action type and rely on force unsubscribe.
5. Re-write the queue to handle this. Also try and reduce calls that might happen because we get a modify patch when in the unsubscribed state

### DIFF
--- a/src/openapi/streaming/subscription.js
+++ b/src/openapi/streaming/subscription.js
@@ -7,7 +7,6 @@ import log from '../../log';
 import {
     ACTION_SUBSCRIBE,
     ACTION_UNSUBSCRIBE,
-    ACTION_MODIFY_SUBSCRIBE,
     ACTION_MODIFY_PATCH,
     ACTION_UNSUBSCRIBE_BY_TAG_PENDING,
 } from './subscription-actions';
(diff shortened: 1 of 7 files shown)


src/openapi/streaming/subscription.js
```diff
@@ -763,31 +762,44 @@ Subscription.prototype.processSnapshot = function(response) {
...
  */
 Subscription.prototype.reset = function() {
     switch (this.currentState) {
         case this.STATE_UNSUBSCRIBED:
         case this.STATE_UNSUBSCRIBE_REQUESTED:
-            // do not do anything if we are on our way to unsubscribed unless the next action would be to subscribe
-            if (this.queue.peekAction() & ACTION_SUBSCRIBE) {
-                break;
-            }
+            // do not do anything - even if the next action is to subscribe, we can go ahead and do that when the unsubscribe response comes back
```
michax, line 778: > even if the next action is to subscribe, we can go ahead and do that when the unsubscribe response comes back

How will scenario with next subscribe know that queue should be reset? Wasn't previous solution clearing the queue if we knew that next action was an subscribe and we are now missing this functionality?

michax (changes requested):

src/openapi/streaming/subscription.js
lukeapage: If we have requested a unsubscribe our job is done - the subscription is on its way to being unsubscribed. If the next action is subscribe- that’s ok because it hasn’t hit the server yet so it can’t possibly be included in what needs to be reset.
What we did before meant that when we get 10 reset all in a row, we started the subscription 10 times.

michax (approved):

PR #334, opened 2021-02-26
Fix streaming issues

yugandhar02:
- Handle disconnect by logging user off
- Handle reconnect and service switching by unsubscribing all subscriptions, stop streaming, start and then re-subscribe
- Check for token expiry during reconnection
- Add more logs and guards

### DIFF
--- a/src/openapi/streaming/connection/connection.js
+++ b/src/openapi/streaming/connection/connection.js
@@ -158,10 +158,11 @@ function Connection(options, baseUrl, failCallback = NOOP) {
         // No next transport available. Report total failure.
         log.error(
             LOG_AREA,
-            'Supported Transport not found.',
+            'Unable to setup initial transport. Supported Transport not found.',
             getLogDetails.call(this),
         );
-        throw new Error('Unable to setup initial transport.');
+
+        failCallback();
     } else {
         log.debug(LOG_AREA, 'Supported Transport found', {
             name: this.transport.name,
--- a/src/openapi/streaming/connection/connection.js
+++ b/src/openapi/streaming/connection/connection.js
@@ -60,7 +60,11 @@ function onTransportFail(error) {
     this.transport.setConnectionSlowCallback(this.connectionSlowCallback);
 
     if (this.state === STATE_STARTED) {
-        this.transport.updateQuery(this.authToken, this.contextId);
+        this.transport.updateQuery(
+            this.authToken,
+            this.contextId,
+            this.authExpiry,
+        );
         this.transport.start(this.options, this.startCallback);
     }
 }
@@ -123,6 +127,7 @@ function Connection(options, baseUrl, failCallback = NOOP) {
     this.baseUrl = baseUrl;
     this.options = options;
     this.authToken = null;
+    this.authExpiry = null;
     this.contextId = null;
     this.transports = getSupportedTransports.call(
         this,
@@ -198,10 +203,12 @@ Connection.prototype.stop = function() {
 Connection.prototype.updateQuery = function(
     authToken,
     contextId,
+    authExpiry,
     forceAuth = false,
 ) {
     this.authToken = authToken;
     this.contextId = contextId;
+    this.authExpiry = authExpiry;
 
     log.debug(LOG_AREA, 'Connection update query', {
         contextId,
@@ -209,7 +216,12 @@ Connection.prototype.updateQuery = function(
     });
 
     if (this.transport) {
-        this.transport.updateQuery(this.authToken, this.contextId, forceAuth);
+        this.transport.updateQuery(
+            this.authToken,
+            this.contextId,
+            this.authExpiry,
+            forceAuth,
+        );
     }
 };
 
--- a/src/openapi/streaming/connection/connection.js
+++ b/src/openapi/streaming/connection/connection.js
@@ -21,6 +21,7 @@ const NOOP = () => {};
 const STATE_CREATED = 'connection-state-created';
 const STATE_STARTED = 'connection-state-started';
 const STATE_STOPPED = 'connection-state-stopped';
+const STATE_DISPOSED = 'connection-state-disposed';
 
 function getLogDetails() {
     return {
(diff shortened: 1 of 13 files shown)


src/openapi/streaming/connection/transport/signalr-core-transport.js
```diff
@@ -24,15 +24,36 @@ function handleLog(level, message) {
     log.warn(LOG_AREA, message);
 }
 
-function buildConnection({ baseUrl, contextId, accessTokenFactory, protocol }) {
+function getRetryPolicy() {
+    return {
+        nextRetryDelayInMilliseconds(retryContext) {
+            if (this.authExpiry < Date.now()) {
```
yugandhar02, line 30: we could also check for error sent by server

src/openapi/streaming/connection/transport/signalr-core-transport.js
```diff
@@ -358,14 +385,20 @@ SignalrCoreTransport.prototype.handleMessageStreamError = function(error) {
 SignalrCoreTransport.prototype.updateQuery = function(
     authToken,
     contextId,
+    authExpiry,
     forceAuth = false,
 ) {
+    if (!this.connection) {
+        return;
+    }
```
yugandhar02, line 393: will remove this check
we first set token then create a connection

src/openapi/streaming/connection/transport/signalr-core-transport.js
```diff
@@ -4,7 +4,7 @@ import * as constants from '../constants';
 
 const LOG_AREA = 'SignalrCoreTransport';
 const NOOP = () => {};
-const RECONNECT_DELAYS = [0, 2000, 3000, 5000, 10000, 20000];
+const RECONNECT_DELAYS = [0, 2000, 3000, 5000, 10000, null];
```
lukeapage, line 7: null means stop trying? what happens if null isn't present? I think this deserves a comment

src/openapi/streaming/connection/transport/signalr-core-transport.js
```diff
@@ -24,15 +24,46 @@ function handleLog(level, message) {
     log.warn(LOG_AREA, message);
 }
 
-function buildConnection({ baseUrl, contextId, accessTokenFactory, protocol }) {
+function getRetryPolicy(transport) {
+    return {
+        nextRetryDelayInMilliseconds(retryContext) {
+            if (transport.authExpiry < Date.now()) {
+                log.warn(LOG_AREA, 'Token expired while trying to reconnect');
```
lukeapage, line 31: its fine that this is a warn to start with. later on we might take it to an info, since it is a normal, ok scenario

src/openapi/streaming/streaming.js
```diff
@@ -174,12 +189,18 @@ function retryConnection() {
...
     this.connectionState = nextState;
 
-    const connectionTransport = this.connection.getTransport();
-    log.debug(LOG_AREA, 'Connection state changed', {
-        changedTo: this.READABLE_CONNECTION_STATE_MAP[this.connectionState],
-        mechanism: connectionTransport && connectionTransport.name,
-        reconnecting: this.reconnecting,
-    });
+    const connectionTransport = this.getActiveTransportName();
+    log.info(
```
lukeapage, line 193: This is going to generate alot of logs - maybe too many. can we just log within signalr core? or make persist depend on signalrcore being enabled for now or something?

src/openapi/streaming/streaming.js
```diff
@@ -221,13 +238,22 @@ function onConnectionStateChanged(nextState) {
...
-            if (this.reconnecting || this.isReset) {
+            if (this.reconnecting) {
                 resetSubscriptions.call(this, this.subscriptions);
                 this.reconnecting = false;
+            }
+
+            // During streaming service switching or reconnect, wait for all subscriptions to unsubscribe before resubscribing again
+            // to avoid hitting throttling limit on subscriptions
+            if (this.isReset) {
+                this.allUnsubscribePendingPromise.then(() => {
```
lukeapage, line 251: I'm wondering if we should instead merge this:
https://github.com/SaxoBank/openapi-clientlib-js/pull/323

This will ensure that if we call reset that (1) we unsubscribe (2) we do not re-subscribe until the unsubscribe is done.

src/openapi/streaming/streaming.js
```diff
@@ -445,6 +477,53 @@ function handleControlMessageResetSubscriptions(referenceIdList) {
...
+        },
+    );
+
+    this.isReset = true;
+
+    this.allUnsubscribePendingPromise = getAllUnsubscribePendingPromise(
+        this.subscriptions,
+    );
+
+    this.disconnect();
```
lukeapage, line 524: and disconnect will always reconnect after it has finished disconnecting?

src/openapi/streaming/streaming.js
lukeapage: Having gone a bit further, I'd rather undo everything you've changed about waiting for unsubscribes and instead call reset after having newly connected.

src/openapi/streaming/connection/connection.js
```diff
@@ -60,7 +74,11 @@ function onTransportFail(error) {
     this.transport.setConnectionSlowCallback(this.connectionSlowCallback);
 
     if (this.state === STATE_STARTED) {
-        this.transport.updateQuery(this.authToken, this.contextId);
+        this.transport.updateQuery(
```
lukeapage, line 77: no changes are needed in legacy signalr? or did I miss it?

yugandhar02: legacy signalr only use first 2 params

src/openapi/streaming/connection/transport/signalr-core-transport.js
yugandhar02: sure.. will add comment

src/openapi/streaming/streaming.js
yugandhar02: ok
will persist depend on signalrcore being enabled

src/openapi/streaming/streaming.js
yugandhar02: sounds good to me

src/openapi/streaming/streaming.js
yugandhar02: yes.. unless we set dispose flag to true

src/openapi/streaming/subscription.js
```diff
@@ -994,6 +994,13 @@ Subscription.prototype.onUnsubscribeByTagComplete = function() {
     onReadyToPerformNextAction.call(this);
 };
 
+/**
```
yugandhar02, line 997: not needed any more

lukeapage (approved):

PR #340, opened 2021-03-05
Add streaming pause, resume support to freeze subscriptions 

yugandhar02:


src/openapi/streaming/streaming.js
```diff
@@ -933,6 +934,36 @@ Streaming.prototype.disconnect = function() {
...
+        this.reconnecting = false;
+        this.retryCount = 0;
+    }
+
+    this.orphanFinder.stop();
+
+    for (let i = 0; i < this.subscriptions.length; i++) {
+        const subscription = this.subscriptions[i];
+        // Mark the subscription as not having a connection and reset it so its state becomes unsubscribed
+        subscription.onConnectionUnavailable();
```
lukeapage, line 951: We do not need to unsubscribe everything?

src/openapi/streaming/streaming.js
```diff
@@ -933,6 +934,36 @@ Streaming.prototype.disconnect = function() {
...
+        subscription.reset();
```
lukeapage, line 952: is there a difference between resetting here vs resetting on resume?

src/openapi/streaming/streaming.js
yugandhar02: reset takes care of that
it checks subscriptions status before subscribing / unsubscribing

src/openapi/streaming/streaming.js
yugandhar02: I don't think there is much difference but its cleaner in server perspective  i.e. cleanup and close ?
This will just push next action i.e. subscribe in the queue until the connection is available

src/openapi/streaming/streaming.js
lukeapage: yes but it will unsubscribe when you come to resume.
However, closing the connection should clean up 🤞 

I wonder if we swap these round - the unsubscribe will be kicked off as it is sync, but further commands like subscribe will not.

src/openapi/streaming/streaming.js
lukeapage: see comment above now

yugandhar02: ahh right

lukeapage (approved):

PR #391, opened 2021-05-05
subcribe before streaming is setup

himanshumeena:
- As of now we queue subscription requests waiting for streaming and only when streaming setup is done we allow it to subscribe.
- setting up streaming connection takes time because of which our subscription also have to wait for it.
- this PR adds an option to pass a Boolean flag **shouldSubscribeBeforeStreamingSetup** while creating a new subscription. 
- when this flag is passed true we will not wait for streaming connection to setup and will subscribe right away. by doing this we can get the initial snapshot response which we can use to render our UI as early as possible.
- later updates will be passed via the streaming channel when its ready

### DIFF
--- a/src/openapi/streaming/subscription.js
+++ b/src/openapi/streaming/subscription.js
@@ -149,6 +149,15 @@ function tryPerformAction(action, args) {
     }
 
     if (
+        this.shouldSubscribeBeforeStreamingSetup &&
+        action === ACTION_SUBSCRIBE &&
+        !this.connectionAvailable
+    ) {
+        performAction.call(this, { action, args });
+        return;
+    }
+
+    if (
         !this.connectionAvailable ||
         this.TRANSITIONING_STATES & this.currentState
     ) {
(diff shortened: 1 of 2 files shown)


src/openapi/streaming/subscription.js
```diff
@@ -684,6 +695,7 @@ Subscription.prototype.STATE_SUBSCRIBED = 0x2;
 Subscription.prototype.STATE_UNSUBSCRIBE_REQUESTED = 0x4;
 Subscription.prototype.STATE_UNSUBSCRIBED = 0x8;
 Subscription.prototype.STATE_PATCH_REQUESTED = 0x10;
+Subscription.prototype.STATE_READY_FOR_UNSUBSCRIBE_BY_TAG = 0x20;
```
himanshumeena, line 698: i think this removed by mistake in this [PR](https://github.com/SaxoBank/openapi-clientlib-js/pull/99)

hiredgun: Yeap, I came to the same conclusion during the TS migration.

hiredgun:
LGTM but let's ask @yugandhar02 or @michax for an additional review as they have more experience with the lib just to be sure that we haven't missed any scenario.

PR #404, opened 2021-05-25
Migrate code to TypeScript

hiredgun:
This PR:
- rewrites the entire repository to TypeScript
- replaces most of the older JS features with the newer API
- removes compilation step with rollup.js since from now on, we will provide not compiled version of the code (we use the emmit feature provided by TS compiler)
- changes JSdoc to TSdoc and removes docs generation based on JSDoc
- improves tools configuration
- removes `libs` folder with `jquery.signalR` 
- improves some tests and adds a new one

Further steps:
- port recent PR from the master with abort signal feature to this branch
- publish beta version and perform final tests in SaxoTrader


src/openapi/streaming/connection/connection.ts
```diff
@@ -0,0 +1,361 @@
...
+import SignalrTransport from './transport/signalr-transport';
+import SignalrCoreTransport from './transport/signalr-core-transport';
+import type { TransportTypes, ConnectionOptions } from '../types';
+import type {
+    StreamingTransportInterface,
+    ReceiveCallback,
+    StateChangeCallback,
+} from './types';
+
+type Callback = (...args: any[]) => unknown | void;
```
lukeapage, line 13: why the unknown?

src/openapi/streaming/connection/connection.ts
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    onOrphanFound() {
+        if (this.transport?.onOrphanFound) {
+            this.transport.onOrphanFound();
+        }
+    }
+
+    onSubscribeNetworkError() {
+        if (this.transport && this.transport.onSubscribeNetworkError) {
```
lukeapage, line 337: nitpick - but 3 different patterns here for the exact same thing - call a function on a object where object or function may be undefined. Would be nice to be consistent.

src/openapi/streaming/connection/constants.ts
```diff
@@ -1,55 +1,54 @@
 /**
  * Event that occurs when the connection state changes.
  */
-export const EVENT_CONNECTION_STATE_CHANGED = 'connectionStateChanged';
+export const EVENT_CONNECTION_STATE_CHANGED = 'connectionStateChanged' as const;
```
lukeapage, line 4: I thought you only need as const on a object property, not on a const export

src/openapi/streaming/connection/transport/signalr-core-transport.ts
```diff
@@ -0,0 +1,623 @@
...
+} from '../types';
+import type SignalR from '@microsoft/signalr';
+
+declare global {
+    interface Window {
+        signalrCore: typeof SignalR;
+    }
+}
+
+type Callback = (...args: any[]) => any;
```
lukeapage, line 22: return any or void?
The use of generic callbacks - was that a time issue, rather than typing each callback?

src/openapi/streaming/connection/transport/signalr-core-transport.ts
```diff
@@ -0,0 +1,623 @@
...
+    state: ConnectionState = constants.CONNECTION_STATE_DISCONNECTED;
+
+    stateChangedCallback: (state: ConnectionState) => void = NOOP;
+    receivedCallback: (data: StreamingMessage) => void = NOOP;
+    errorCallback = NOOP;
+    unauthorizedCallback = NOOP;
+    setConnectionSlowCallback = NOOP;
+    transportFailCallback;
+
+    utf8Decoder!: TextDecoder;
```
lukeapage, line 64: Is this exclamation mark to tell typescript it won't be null?

src/openapi/streaming/connection/transport/signalr-core-transport.ts
```diff
@@ -0,0 +1,623 @@
...
+    }
+
+    static isSupported = function () {
+        return (
+            window.signalrCore &&
+            typeof window.signalrCore.HubConnectionBuilder === 'function' &&
+            typeof window.Uint8Array === 'function' &&
+            typeof window.TextDecoder === 'function' &&
+            // This check can be removed once signalr team resolves below issue
+            // https://github.com/dotnet/aspnetcore/issues/29424
```
lukeapage, line 93: given the outcome here: https://github.com/dotnet/aspnetcore/issues/29424

maybe just remove the comment?

src/openapi/streaming/connection/transport/signalr-transport.ts
```diff
@@ -0,0 +1,137 @@
+import log from '../../../../log';
+import * as transportTypes from '../transportTypes';
+import * as constants from '../constants';
+import type { StreamingMessage } from '../../types';
+import type { StreamingTransportInterface } from '../types';
+
+const LOG_AREA = 'SignalRTransport';
+const NOOP = () => {};
+
+type Callback = (...args: any) => any;
```
lukeapage, line 10: another Callback type with any return

src/openapi/streaming/connection/transportTypes.ts
```diff
@@ -0,0 +1,7 @@
+export const PLAIN_WEBSOCKETS = 'plainWebSockets' as const;
```
lukeapage, line 1: unnecessary as const (I think)

src/openapi/streaming/parser/wrappers/protobuf-wrappers.ts
```diff
@@ -0,0 +1,33 @@
...
+                        seconds?: number;
+                    };
+
+                    // Date with support for nano precision
+                    const date = new Date(
+                        Number(seconds) * 1000 +
+                            Math.floor(Number(nanos) / 1000000),
+                    );
+
+                    // @ts-expect-error FIXME TS is complaining that a string is returned,
```
lukeapage, line 24: Looks like something we intended to fix before v9 release

src/openapi/streaming/streaming.ts
```diff
@@ -0,0 +1,1124 @@
...
+            // in case the refresh timer has disappeared, ensure authProvider is
+            // fetching a new token
+            const transport = this.getActiveTransportName();
+            this.authProvider.refreshOpenApiToken();
+            this.authProvider.one(
+                this.authProvider.EVENT_TOKEN_RECEIVED,
+                () => {
+                    // FIXME looks like this block is currently never executed since
+                    // isReconnection arg is never provided
+                    if (isReconnection && !this.reconnecting) {
```
lukeapage, line 307: Fix me.

src/openapi/streaming/streaming.ts
```diff
@@ -0,0 +1,1124 @@
...
+
+    /**
+     * Fires heartbeats to relevant subscriptions
+     * @param heartbeatList - heartbeatList
+     */
+    private handleControlMessageFireHeartbeats(
+        heartbeatList: types.Heartbeats[] | null,
+    ) {
+        log.debug(LOG_AREA, 'heartbeats received', { heartbeatList });
+        // @ts-expect-error FIXME heartbeatList may be null, decide whether to throw or use an empty array as a default
```
lukeapage, line 611: fix me

src/openapi/streaming/streaming.ts
```diff
@@ -0,0 +1,1124 @@
...
+        for (let i = 0; i < heartbeatList.length; i++) {
+            // @ts-expect-error
```
lukeapage, line 613: why expect an error?

src/openapi/streaming/subscription-actions.ts
```diff
@@ -0,0 +1,12 @@
+export const ACTION_SUBSCRIBE = 0x1 as const;
```
lukeapage, line 1: unnecessary as const?

src/openapi/streaming/subscription.ts
```diff
@@ -0,0 +1,1160 @@
...
+            // - (reset by heartbeat lapse) - this indicates a bug in the library since this shouldn't happen
+            //   - in this case the contextId should be the same and we will probably get messages that cannot be matched to a subscription
+            return;
+        }
+
+        this.setState(this.STATE_SUBSCRIBED);
+
+        this.inactivityTimeout = responseData.InactivityTimeout || 0;
+
+        // @ts-expect-error FIXME this always evaluates to false - find out what to do with this
```
lukeapage, line 534: FIXME

!responseData.InactivityTimeout === 0
-->
responseData.InactivityTimeout === 0

src/openapi/streaming/subscription.ts
```diff
@@ -0,0 +1,1160 @@
...
+        ) {
+            return Infinity;
+        }
+
+        // Follows the same pattern as the old library, not giving any grace period for receiving a heartbeat
+        // if it was required, it could be added on here
+
+        // @ts-expect-error FIXME this.latestActivity  may be undefined since it was not
+        // initialized in the old code - verify whether it can be undefined at this point or initialized with 0 in  the constructor
+        const diff = now - this.latestActivity;
```
lukeapage, line 1152: not sure it can happen but seems safe to initialise it. maybe to Date.now? that makes the diff small which makes the return value larger which is correct - at start time till orphaned should be big

src/openapi/transport/putPatchDiagnosticsQueue.ts
```diff
@@ -0,0 +1,113 @@
...
+import type TransportCore from './core';
+import type {
+    HTTPMethodType,
+    HTTPMethod,
+    StringTemplateArgs,
+    RequestOptions,
+} from '../../types';
+import type { ITransport } from './transport-base';
+
+// fix-me typo
```
lukeapage, line 12: Its ok to just fix I think

src/openapi/transport/queue.ts
```diff
@@ -0,0 +1,182 @@
...
+        this.queue.length = 0;
+    }
+
+    protected runQueueItem(item: QueueItem) {
+        this.transport[item.method](...item.args).then(
+            (...args: any[]) => {
+                item.resolve(...args);
+            },
+            // @ts-expect-error as it should return promise but returning direct void
+            (result: { status: HTTPStatusCode }, ...args: [any]) => {
```
lukeapage, line 142: nitpick but should it be ReadonlyArray<any> ? currently it says there will only ever be one argument

src/openapi/transport/types.ts
```diff
@@ -0,0 +1,40 @@
+import type { RequestOptions, StringTemplateArgs } from '../../types';
+
+export type HTTPStatusCode = 401 | 404 | 200 | 201 | 500;
```
lukeapage, line 3: there are more status codes we use ;- looks a bit strange, and does it mean as a consumer if I ant to check if we got 429 that it will give a ts error.

I wonder if it is not worth it and we should type as number

src/utils/types.ts
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable max-len */
```
lukeapage, line 1: shouldn't be needed? we use prettier so it should be disabled?

lukeapage (changes requested):
Fantastic work! Am very excited this is done and so much cleaner now.

src/openapi/authProvider.ts
```diff
@@ -0,0 +1,465 @@
...
+        this.retries = 0;
+
+        this.createTimerForNextToken();
+    }
+
+    /**
+     * This internal method refreshes the token no matter what and should only
+     * be called if we know we are in the correct state to do it.
+     */
+    private refreshToken = () => {
```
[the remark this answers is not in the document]
hiredgun: yes, the arrow is used since it's passed to setTimeout. I used constructor binding in one place since I noticed that IDE provides better params/doc hints for methods but in this case, it doesn't matter.

src/openapi/streaming/connection/constants.ts
hiredgun: This was added since TS converts consts assigned to a class property as a string. I realized that checking it in [this playground ](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAwgeQHJIKJwCoElkH0DKGAghqrlkltkQDJYBaFA4jALwwDk0ATgKYCGAWwCWYAOYARYRFBgwvYFABKvAI4BXXtF4ATDgG4AsACgTAegBUFkzAswAsvzD8xWmBHUAjGd2EAHKGFwCBgoEFCAC14YBD9eMBgiAAUsdyg+IVExd15uADdhYF4AOhs7BDAimGAMqF1I6RgAd2EAG1aYYUFBXWF+OtaATxh+usEA0PDofm5YKCi0jJFxHPzC3jKzE2BW-ggQ-HSBZeyAbxsYS7MzGAydcCH4ZDRMHCQCYlJySmo6RiQWOxECh0Ng8IQSGQKFQsLQGMwjMZLpdgS8we8IV9ob94QC2E8Qa9wZ8oT9YX8ETBrjAADx0mAAWiZzJZzOaUQSTWiewgwjECX41WCUAANDAMPgYIJ+ABrNzCWCCnhZEwXS7pPmubgACmcPQAXJweMcslIZOB5IoVBotHU9DAAD6cABmIBAHAAlDBzqYkciAL6qv3q23ar0+5GRhoQYoasRa7XzaTFVGgt4fSHfGFw-5MD2IyOB4xFoPUgBqkqDzvUlUC4FCvnjuV1Ql4hq4R0y4jNsktyjUmm09qdHFd7vDQZLxjjCdTRIxJKzONz+aAA). An alternative would be to remove `as const` and add `readonly` modifier to each constant class property.

src/openapi/streaming/connection/transport/signalr-core-transport.ts
hiredgun: Yes, in this case, it's initialized in the constructor. If there's no TextDecoder constructor on the window object, it actually won't be initialized but we call transportFailCallback and assume this case is handled.

src/openapi/streaming/connection/connection.ts
hiredgun: Not needed, looks like a leftover from some fixes, will check if this type fits all usages.

src/openapi/streaming/connection/connection.ts
hiredgun: Fixed

src/openapi/streaming/streaming.ts
hiredgun: Checked and `isReconnection` is provided in one case. Will remove the comment.

src/openapi/streaming/connection/transport/signalr-transport.ts
hiredgun: Fixed

src/openapi/streaming/connection/connection.ts
hiredgun: Fixed

src/openapi/streaming/connection/transport/signalr-core-transport.ts
hiredgun: I cleaned up the callback mess. Thanks for pointing this out!

src/openapi/streaming/streaming.ts
hiredgun: connection with the suppression above, fixed

src/openapi/streaming/streaming.ts
hiredgun: fixed

src/openapi/transport/putPatchDiagnosticsQueue.ts
hiredgun: Accidentally fixed it while I was doing a global replacement - removed the comment

src/openapi/transport/queue.ts
hiredgun: Looks like a mistake. Changed it to an array.

src/openapi/transport/types.ts
hiredgun: yeah :D changed it to number

src/utils/types.ts
hiredgun: yeap, removed the eslint rule

src/openapi/streaming/subscription.ts
hiredgun: this method is used in `orphaFinder`, and with the current implementation, in case `latestActivity` or `inactivityTimeout` is `undefined`, the `NaN` is returned:
```ts
const timeTillOrphaned = subscription.timeTillOrphaned(now);
if (timeTillOrphaned <= 0) {
    orphanedSubscriptions.push(subscription);
} else if (timeTillOrphaned < newNextUpdateIn) {
    foundNextUpdate = true;
    newNextUpdateIn = timeTillOrphaned;
}
```

 so if we return `Infinity` the same behavior should be preserved. I made it like this in the latest commit. Let me know if there might be an issue with this.

src/openapi/streaming/parser/wrappers/protobuf-wrappers.ts
hiredgun: Discussed it with @michax and this is expected. I updated the comment.

src/number-formatting/format.ts
```diff
@@ -1,38 +1,36 @@
...
+ * @param precision - precision
  */
-function convertNumbertToString(number, precision) {
+function convertNumberToString(number: number, precision: number) {
     let numberString = String(number);
 
     // if the number is in scientific notation, convert to decimal
     if (/\d+\.?\d*e[+-]*\d+/i.test(numberString)) {
-        numberString = number.toFixed(precision).trim('0');
+        numberString = number.toFixed(precision).trim();
```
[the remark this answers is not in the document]
lukeapage: aha. yes lets remove the trim.

src/openapi/streaming/connection/constants.ts
lukeapage: I see so it is to avoid `as const` where we use them in the class?

lukeapage: ok, looked at the playground. Weird!

lukeapage (approved):

PR #410, opened 2021-05-28
Migrate files to TypeScript with preserved history

hiredgun:
### READ BEFORE MERGE!
We should NOT squash merge this PR since this will not preserve the rename commit and we will lose git history on migrated files. We SHOULD either use `create and merge` or `rebase and merge` from the merge dropdown while finishing this PR.

The intent of this PR is to preserve git history after migration to TS (#404). This is because, when you rename a file and immediately add a lot of changes, git thinks the original file is deleted and another one is created. In order to preserve the history we should rename a file, commit and only after that add new changes - this was not done on the migration branch. 

Solution:
- this branch was created from master
- all js files were renamed to ts and committed
- squash merge commits (#404 and #408) from develop branch was cherry-picked to this branch

When you check git history of a renamed file, by default, `git log ./path/to/file` command displays only history from a `rename` commit. In order to see the entire history you need to use `--follow` command e.g: `git log --follow /path/to/file`. From what I checked WebStorm can see entire history so probably it uses `--follow` by default.


hiredgun:
Merged one additional PR (#412) to this one which disables "Connection state change" message persistence

michax (approved):

PR #416, opened 2021-06-03
websocket transport streaming heartbeat added

himanshumeena:
- Due to the fact that browsers may disconnect web-socket connection without knowing so, we need to add a heartbeat protocol on top of web-socket in the web-socket streaming service.

- To enable heartbeats on a connection (or reconnection) the following query parameter must be added to the connection URL: **&sendHeartbeats=true**

- The heartbeat messages follow the usual control message format, so they come with the reference id **_connectionheartbeat**

-  So by sending the "sendHeartbeats=true" as queryParam a connection heartbeat is sent after 2 seconds of inactivity on a streaming connection, i.e. when there has been no heartbeat or other streaming message for 2 seconds

- In this change we try to reconnect the web socket if we don't receive any message on the web socket for 2.5 seconds and we are not waiting for authorization with new token

### DIFF
--- a/src/openapi/streaming/connection/connection.ts
+++ b/src/openapi/streaming/connection/connection.ts
@@ -200,16 +200,7 @@ class Connection {
             return this.createTransport(baseUrl);
         }
 
-        let options;
-        if (this.options && this.options.isWebsocketStreamingHeartBeatEnabled) {
-            if (SelectedTransport.name === 'WebsocketTransport') {
-                options = {
-                    isWebsocketStreamingHeartBeatEnabled: true,
-                };
-            }
-        }
-
-        return new SelectedTransport(baseUrl, this.onTransportFail, options);
+        return new SelectedTransport(baseUrl, this.onTransportFail);
     }
 
     private getSupportedTransports(
--- a/src/openapi/streaming/connection/connection.ts
+++ b/src/openapi/streaming/connection/connection.ts
@@ -200,7 +200,16 @@ class Connection {
             return this.createTransport(baseUrl);
         }
 
-        return new SelectedTransport(baseUrl, this.onTransportFail);
+        let options;
+        if (this.options && this.options.isWebsocketStreamingHeartBeatEnabled) {
+            if (SelectedTransport.name === 'WebsocketTransport') {
+                options = {
+                    isWebsocketStreamingHeartBeatEnabled: true,
+                };
+            }
+        }
+
+        return new SelectedTransport(baseUrl, this.onTransportFail, options);
     }
 
     private getSupportedTransports(
--- a/src/openapi/streaming/connection/transport/websocket-transport.spec.ts
+++ b/src/openapi/streaming/connection/transport/websocket-transport.spec.ts
@@ -498,4 +498,69 @@ describe('openapi WebSocket Transport', () => {
             });
         });
     });
+
+    describe('websocket inactivity finder', (done) => {
+        it('start and stop should work correctly', () => {
+            const transport = new WebSocketTransport(BASE_URL, undefined, {
+                isWebsocketStreamingHeartBeatEnabled: true,
+            });
+            transport.updateQuery(AUTH_TOKEN, CONTEXT_ID);
+            transport.start({}, () => {});
+            fetchMock.resolve(200, {});
+
+            transport.authorizePromise.then(() => {
+                // on ws open should start
+                transport.socket.readyState = 1; // WebSocket internal state equal open
+                transport.socket.onopen();
+
+                expect(transport.inactivityFinderEnabled).toBe(true);
+                expect(
+                    transport.inactivityFinderNextUpdateTimeoutId,
+                ).to.not.equal(null);
+
+                // on ws close should stop
+                transport.socket.readyState = 3; // WebSocket internal state equal closed
+                transport.socket.onclose({ code: 1001 });
+
+                expect(transport.inactivityFinderEnabled).toBe(null);
+                expect(transport.inactivityFinderNextUpdateTimeoutId).to.equal(
+                    null,
+                );
+                done();
+            });
+        });
+
+        it('should reconnect if connection is established and there is no messages for more than 2.5 seconds', (done) => {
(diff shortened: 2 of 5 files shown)


src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -377,6 +391,43 @@ class WebsocketTransport implements StreamingTransportInterface {
         }
     }
 
+    private startInactivityFinder = () => {
```
lukeapage, line 394: is there a chance of start being called when it is already running?

src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -340,6 +353,7 @@ class WebsocketTransport implements StreamingTransportInterface {
         socket.onmessage = null;
         socket.onclose = null;
         socket.close(socketCloseCodes.NORMAL_CLOSURE, CLOSE_REASON_DESTROY);
+        this.stopInactivityFinder();
```
lukeapage, line 356: we always call destroy socket right - even when the socket has closed itself?

src/openapi/streaming/connection/connection.ts
```diff
@@ -200,7 +200,16 @@ class Connection {
             return this.createTransport(baseUrl);
         }
 
-        return new SelectedTransport(baseUrl, this.onTransportFail);
+        let options;
+        if (this.options && this.options.isWebsocketStreamingHeartBeatEnabled) {
```
lukeapage, line 204: there is already a way of passing options - on start - would it be neater to do it in the same way? we don't need the option until it is started..

lukeapage (changes requested):

src/openapi/streaming/connection/transport/websocket-transport.ts
himanshumeena: yes we do call socket destroy when socket close itself
i have checked the logout, reconnect and network error scenarios

src/openapi/streaming/connection/transport/websocket-transport.ts
himanshumeena: startInactivityFinder is attached to socket.onopen listener and it will be invoked only once when we create the stocket, i have added logging to track this further

src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -377,6 +391,43 @@ class WebsocketTransport implements StreamingTransportInterface {
         }
     }
 
+    private startInactivityFinder = () => {
+        this.inactivityFinderEnabled = true;
```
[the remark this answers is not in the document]
himanshumeena: changed to "inactivityFinderRunning" as suggested

src/openapi/streaming/connection/connection.ts
himanshumeena: using start method now to set the flag

src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -377,6 +390,53 @@ class WebsocketTransport implements StreamingTransportInterface {
...
+                LOG_AREA,
+                'Inactivity finder reconnecting since no message recieved since 3 seconds',
+            );
+            this.reconnect();
+        }
+
+        this.inactivityFinderNextUpdateTimeoutId = window.setTimeout(
+            this.onInactivityFinderUpdate,
+            MAX_INACTIVITY_WAIT_TIME,
+        );
```
hiredgun, line 437: One thought here,  seems like the `lastMessageTime` is not reset anywhere, let's say that we detected inactivity so we reconnect and schedule next check after 3s. I'm wondering if it is possible to jump to this method again after reconnection without `lastMessageTime` being updated which would mean we reconnect immediately again.

src/openapi/streaming/connection/transport/websocket-transport.spec.ts
```diff
@@ -498,4 +498,79 @@ describe('openapi WebSocket Transport', () => {
...
+                    new ArrayBuffer(dataBuffer.length + 17),
+                );
+                payload.set(
+                    [3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 56, 0, 134, 12, 0, 0],
+                    0,
+                );
+                payload.set(dataBuffer, 17);
+                transport.socket?.onmessage?.({ data: payload.buffer } as any);
+
+                tick(3001);
```
hiredgun, line 567: what about checking the state before 3s to make sure it actually has changed?

src/openapi/streaming/connection/transport/websocket-transport.ts
himanshumeena: `lastMessageTime` only updates when we get snapshot, deltas or heartbeat message.
and yes if we still not received any message during waiting for 3 seconds it should reconnect after that and its the expected behavior

src/openapi/streaming/connection/transport/websocket-transport.spec.ts
himanshumeena: asserted for previous state before 3seconds

src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -377,6 +390,53 @@ class WebsocketTransport implements StreamingTransportInterface {
         }
     }
 
+    private startInactivityFinder = () => {
+        if (this.inactivityFinderRunning) {
+            log.warn(
+                LOG_AREA,
+                'Starting inactivityFinder when already started',
+            );
```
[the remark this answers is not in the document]
himanshumeena: yep we can 
returning early now

src/openapi/streaming/connection/transport/websocket-transport.ts
hiredgun: yeah, but in case the lastMessageTime is not updated by any snapshot/heartbeat etc. after reconnection, we will immediately reconnect since we will have the old lastMessageTime - ofc only if such a scenario is even possible.

hiredgun (approved):

PR #432, opened 2021-06-10
Enhanced getActualDecimals function 

harshitsilly:
Why?
`formatNoRounding` function is appending 0 values(uptil 8 decimal) for large numbers. 
- Reason - `getActualDecimals`  is returning incorrect decimal length due to `number-Math.float(number)`  and `toFixed` which  approximate value due to floating precision 
- In case of  `formatNoRounding(92893282191.78,2)`, getActualDecimal will return the decimal part `"0.77999878"` as toFixed(8) is not able to roundoff (Thus length 8) and we get `'92893282191.78000000'` which is wrong.

Soln
- Will adjust constant 8 that is passed to `toFixed`.
- Browser start rounding off(precision) decimal number after 16-17 chars. So by subtracting const 16(taking assumption) with non-decimal places of number we can get the actual decimal place where the browser is able to round off correctly.

### DIFF
--- a/src/number-formatting/index.ts
+++ b/src/number-formatting/index.ts
@@ -155,7 +155,7 @@ class NumberFormatting implements NumberFormattingOptions {
 
         return (number - Math.floor(number))
             .toFixed(maxDecimals)
-            .substring(2, 10)
+            .substring(2, 2 + maxDecimals)
             .replace(numberOfZerosRx, '').length;
     }
 }
--- a/src/number-formatting/index.ts
+++ b/src/number-formatting/index.ts
@@ -148,6 +148,7 @@ class NumberFormatting implements NumberFormattingOptions {
     /**
      * Returns the actual number of decimals that a number has.
      * @param number - number or numeric string
+     * @param isNoRounding - oprional or boolean
      */
     getActualDecimals(number: number, isNoRounding?: boolean) {
         number = Math.abs(number);
@@ -169,7 +170,7 @@ class NumberFormatting implements NumberFormattingOptions {
                 .length;
         }
 
-        const decimal = number.toString().split('.')[1];
+        const decimal = sNumber.split('.')[1];
         const decimalLength = decimal ? decimal.length : 0;
         return Math.min(decimalLength, 8);
     }
--- a/src/number-formatting/index.ts
+++ b/src/number-formatting/index.ts
@@ -5,7 +5,15 @@ import shortFormat from './short-format';
 
 const numberOfZerosRx = /0+$/;
 
-// browsers [ie,chromium,firefox] start rounding of decimal number after 16-17 chars.
+/**
+ * Come up with this number through experimentation.
+ * In js if the number is greater than 16/17 chars then it starts to round off
+ * @example
+ * let c = 123456789123456789 , c becomes 123456789123456780
+ * 123456789123456.789 becomes 123456789123456.78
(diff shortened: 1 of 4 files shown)


src/number-formatting/index.ts
```diff
@@ -143,13 +148,31 @@ class NumberFormatting implements NumberFormattingOptions {
     /**
      * Returns the actual number of decimals that a number has.
      * @param number - number or numeric string
+     * @param isNoRounding - oprional or boolean
```
lukeapage, line 151: Is no rounding doesn’t make sense as a parameter here - get actual decimals should always return the actual decimals..

lukeapage:
The problem with the existing code is that it assumes that 8 is always going to be a number of decimals that JavaScript can handle with high precision- but because the number is large, that is not the case.

A different approach (not sure if it is better though) would be to adjust the 8 constant to toFixed depending on how many characters the whole number has.

lukeapage:
The point of the toFixed was in the case that we are representing 1.3 and toString returns 1.299999999 - in this case your code wouldn’t work.. though it seems unfortunately we have no test cases for this if the tests pass.

src/number-formatting/index.ts
yugandhar02: > The problem with the existing code is that it assumes that 8 is always going to be a number of decimals that JavaScript can handle with high precision- but because the number is large, that is not the case.
> 
> A different approach (not sure if it is better though) would be to adjust the 8 constant to toFixed depending on how many characters the whole number has.

I think this is something we can do
but what should be total chars we should check against ?
Do you mean to say something like this ?

`
const maxChars = 16;

function getActualDecimals(number) {
        number = Math.abs(number);
        
        let maxDecimals = maxChars  - Math.floor(number).toString().length;
        maxDecimals = Math.max(0, Math.min(maxDecimals, 8))
        
        return (number - Math.floor(number))
            .toFixed(maxDecimals)
            .substring(2, 10)
            .replace(numberOfZerosRx, '').length;
    }
`

harshitsilly:
updated the PR with toFixed logic with variable const.
made and assumption of 16 chars for const maxDecimals.

src/number-formatting/index.ts
```diff
@@ -5,6 +5,9 @@ import shortFormat from './short-format';
 
 const numberOfZerosRx = /0+$/;
 
+// browsers [ie,chromium,firefox] start rounding of decimal number after 16-17 chars.
```
lukeapage, line 8: its not browser specific, I'm pretty sure it is javascript specific.

did you come up with this number through experimentation or some resource? Can you just expand on how this number was arrived at, so that if we need to adjust it we know how we got this?

harshitsilly: We have come up with this number through experimentation.
Simply in console if we have number greater than 16/17 chars  then js starts to round off
for ex- let c = `123456789123456789` , c becomes `123456789123456780`

`123456789123456.789` becomes `123456789123456.78`

`1234567891234567.1` becomes `1234567891234567`
`1234567891234567.8` becomes `1234567891234567.8`


+ if number is greater than 21 than it is converted to exponential notation

cc: @yugandhar02

lukeapage: Ok - please add this to the comment and it’s good to go

harshitsilly: Done

lukeapage (approved):

PR #433, opened 2021-06-10
wait 1 minute when 3 resets received within a minute

himanshumeena:
When we get a reset on the subscription, we re-subscribe.

However we have seen in the past that when open api is in a bad state, they may reset immediately.

So we would like to detect if we get a 3rd reset within 1 minute that we
1) Log an error
2) wait a minute

and obviously if unsubscribe occurs whilst waiting then it shouldn't resume.

### DIFF
--- a/src/openapi/streaming/subscription.spec.ts
+++ b/src/openapi/streaming/subscription.spec.ts
@@ -1708,6 +1708,93 @@ describe('openapi StreamingSubscription', () => {
                 });
             });
         });
+
+        describe('3 resets within 1 minute', () => {
+            it('should unsubscribe immediately if unsubscribe requested while waiting for publisher to respond', (done) => {
+                const subscription = new Subscription(
+                    '123',
+                    transport,
+                    'servicePath',
+                    'src/test/resource',
+                    {},
+                    createdSpy,
+                    { onUpdate: updateSpy },
+                );
+
+                subscription.reset();
+                subscription.reset();
+                subscription.reset();
+
+                expect(subscription.currentState).toEqual(
+                    subscription.STATE_PUBLISHER_DOWN,
+                );
+                expect(
+                    subscription.waitForPublisherToRespondTimer,
+                ).toBeTruthy();
+
+                // should unsubscribe immediately if requested
+                subscription.onUnsubscribe();
+
+                expect(subscription.currentState).toEqual(
+                    subscription.STATE_UNSUBSCRIBE_REQUESTED,
+                );
+                expect(subscription.waitForPublisherToRespondTimer).toBeFalsy();
+
+                done();
+            });
+
+            it('should wait for publisher to respond', (done) => {
+                const subscription = new Subscription(
+                    '123',
+                    transport,
+                    'servicePath',
+                    'src/test/resource',
+                    {},
+                    createdSpy,
+                    { onUpdate: updateSpy },
+                );
+
+                subscription.reset();
+                subscription.reset();
+                subscription.reset();
+
+                expect(subscription.currentState).toEqual(
+                    subscription.STATE_PUBLISHER_DOWN,
+                );
+                expect(
+                    subscription.waitForPublisherToRespondTimer,
+                ).toBeTruthy();
+
+                // should queue actions while waiting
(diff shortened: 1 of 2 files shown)


src/openapi/streaming/streaming.ts
```diff
@@ -738,6 +740,35 @@ class Streaming extends MicroEmitter<EmittedEvents> {
         this.orphanFinder.update();
     }
 
+    /**
+     * If we get 3 resets within 1 minute then we wait for 1 minute
+     * since it may indicate some problem with publishers
+     */
+    private checkIfPublishersDown(resetTimeStamp) {
```
lukeapage, line 747: This is across all subscriptions- I would rather this were per subscription. It might make some of the subscription logic easier too as you can directly put it into a new state.

himanshumeena: changed to per subscription as suggested

src/openapi/streaming/subscription.spec.ts
```diff
@@ -1708,6 +1708,97 @@ describe('openapi StreamingSubscription', () => {
...
+                          "args": undefined,
+                        },
+                      ],
+                    }
+                `);
+                expect(subscription.currentState).toEqual(
+                    subscription.STATE_PUBLISHER_DOWN,
+                );
+
+                setTimeout(() => {
```
hiredgun, line 1784: why setTimeout is used here?

src/openapi/streaming/subscription.ts
```diff
@@ -240,6 +247,33 @@ class Subscription {
...
+            if (
+                !this.waitForPublisherToRespondTimer &&
+                this.resetTimeStamps[2] - this.resetTimeStamps[0] <
+                    MIN_WAIT_FOR_PUBLISHER_TO_RESPOND_MS
+            ) {
+                // 3 reset within 1 minute so wait for 1 minute for pubslisher to respond
+                log.warn(LOG_AREA, 'Received 3 resets within 1 minute');
+                this.setState(this.STATE_PUBLISHER_DOWN);
+
+                this.waitForPublisherToRespondTimer = window.setTimeout(() => {
```
hiredgun, line 268: Shouldn't we empty the resetTimestamps array here too? Seems like once we have 3 timestamps we will perform this check with every reset since the array will always have 4 items? Is that expected? Am I missing sth?

src/openapi/streaming/subscription.ts
```diff
@@ -934,6 +986,10 @@ class Subscription {
                 // We are about to unsubscribe by tag, so no need to do anything
                 return;
 
+            case this.STATE_PUBLISHER_DOWN:
```
hiredgun, line 989: What about including this one in the first case?

src/openapi/streaming/subscription.ts
himanshumeena: - so this function will only be called when reset happens
- we will always have latest 3 reset timestamps in this array
```
if (this.resetTimeStamps.length >= 3) {
            this.resetTimeStamps = this.resetTimeStamps.slice(-3);
```

-  we do wait for 1minute when we get 3resets within a minute and while waiting we don't start another wait 
```
!this.waitForPublisherToRespondTimer &&
                this.resetTimeStamps[2] - this.resetTimeStamps[0] <
                    MIN_WAIT_FOR_PUBLISHER_TO_RESPOND_MS
```
- after that 3 will need 3 more news resets that should happen within one minute to trigger the wait again

src/openapi/streaming/subscription.ts
himanshumeena: took inspiration from `STATE_READY_FOR_UNSUBSCRIBE_BY_TAG` 
that will work but i took it out for more readability, also 
```
case this.STATE_UNSUBSCRIBED:
            case this.STATE_UNSUBSCRIBE_REQUESTED:
```
looks more concerned related to unsubscribe, so thought better to keep it separate.

src/openapi/streaming/subscription.spec.ts
himanshumeena: not needed ,removed

hiredgun (approved):

PR #437, opened 2021-06-16
Pass url details to shouldUseCloud method 

kiermasp:
Add second argument to _shouldUseCloud_ to have more details and decide if specific endpoint should use cloud.

New type:

```
URLDetails = {
      method: string
      servicePath: string
      url: string
  };
```


lukeapage:
Lgtm. Can you update the changelog and package version number?

michax (approved):

PR #446, opened 2021-06-24
Detect multiple orphans and reset the connection

lukeapage:


src/openapi/streaming/streaming.ts
```diff
@@ -701,6 +711,10 @@ class Streaming extends MicroEmitter<EmittedEvents> {
             },
         );
 
+        this.reconnect();
+    }
+
+    reconnect() {
```
lukeapage, line 717: plan is to call reconnect when problems detected, protected by a ff

src/openapi/streaming/streaming.ts
```diff
@@ -744,11 +765,86 @@ class Streaming extends MicroEmitter<EmittedEvents> {
...
+            log.info(LOG_AREA, 'Subscription has become orphaned - resetting', {
+                referenceId: subscription.referenceId,
+                streamingContextId: subscription.streamingContextId,
+                servicePath: subscription.servicePath,
+            });
+
+            const now = Date.now();
+            const lastMinute = now - 60 * 1000;
+            this.orphanEvents = this.orphanEvents.filter(
+                (orphanEvent) => orphanEvent.datetime < lastMinute,
```
hiredgun, line 778: shouldn't this be `datetime > lastMinute` to get only events fired withing last minute? 🤔

PR #448, opened 2021-06-25
Log received message count on disconnect

robatwilliams:
Adding this so we can compare with the streaming gateway's logs which look like this:

```
Abort called on Session. Sent messages: 20496. Received messages: 20495
```

This PR needs further work. @jdhodges

### DIFF
--- a/src/openapi/streaming/streaming.ts
+++ b/src/openapi/streaming/streaming.ts
@@ -404,7 +404,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
             case this.CONNECTION_STATE_DISCONNECTED:
                 log.info(
                     LOG_AREA,
-                    'Connection is disconnected',
+                    'Connection disconnected',
                     { contextMessageCount: this.contextMessageCount },
                     { persist: true },
                 );
@@ -428,7 +428,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
             case this.CONNECTION_STATE_RECONNECTING:
                 log.info(
                     LOG_AREA,
-                    'Connection is reconnecting',
+                    'Connection disconnected. Reconnecting...',
                     { contextMessageCount: this.contextMessageCount },
                     { persist: true },
                 );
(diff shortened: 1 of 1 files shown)


src/openapi/streaming/streaming.ts
```diff
@@ -400,6 +402,13 @@ class Streaming extends MicroEmitter<EmittedEvents> {
 
         switch (this.connectionState) {
             case this.CONNECTION_STATE_DISCONNECTED:
+                log.info(
+                    LOG_AREA,
+                    'Connection disconnected',
+                    { contextMessageCount: this.contextMessageCount },
+                    { persist: true },
```
robatwilliams, line 409: Need to figure out how often this happens and decide if it (and the below call) should be persisted, to avoid spamming ourselves

src/openapi/streaming/streaming.ts
```diff
@@ -417,6 +426,13 @@ class Streaming extends MicroEmitter<EmittedEvents> {
                 break;
 
             case this.CONNECTION_STATE_RECONNECTING:
+                log.info(
+                    LOG_AREA,
+                    'Connection disconnected. Reconnecting...',
```
robatwilliams, line 431: This wording reflects that connections don't get reconnected, you just make a new connection. Reconnecting is just something the Connection class here does

src/openapi/streaming/streaming.ts
```diff
@@ -143,6 +143,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
     };
     reconnecting = false;
     contextId!: string;
+    contextMessageCount!: number;
```
robatwilliams, line 146: Not sure what to call this...

underlying connection message count - what it really is
connection message count - misleading because the Connection class is long-lived
session message count - what OpenAPI calls it

robatwilliams:
Superseded by #450

src/openapi/streaming/streaming.ts
lukeapage: we did persist it until recently. Can we just check if we log it inside saxotrader too.
So I think its ok.

jdhodges: In saxotrader it has some spam mitigation for disconnections since it checks if a timeout is already in progress before logging  - do you think we'll need something similar here or we'll be alright logging on every disconnection?

PR #450, opened 2021-06-28
Message count log

jdhodges:
Copy since I don't have permission to push to @robatwilliams fork #448

Adds message count and last message received datetime on disconnection.

### DIFF
--- a/src/openapi/streaming/streaming.ts
+++ b/src/openapi/streaming/streaming.ts
@@ -144,6 +144,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
     reconnecting = false;
     contextId!: string;
     contextMessageCount!: number;
+    contextLastMessageReceived!: string;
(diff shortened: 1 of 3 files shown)


src/openapi/streaming/streaming.ts
```diff
@@ -492,12 +517,12 @@ class Streaming extends MicroEmitter<EmittedEvents> {
...
-            this.processUpdate(updates);
-            return;
+            updates = [updates];
         }
 
-        for (let i = 0; i < updates.length; i++) {
-            const update = updates[i];
+        for (const update of updates) {
+            this.contextMessageCount++;
+            this.contextLastMessageReceived = new Date().toISOString();
```
lukeapage, line 525: If you rebase from master I've added a lastActivity already for this- can you delete this and use my field?
its just a number but I think thats good to avoid us doing this for every streaming message.

lukeapage (changes requested):

PR #463, opened 2021-07-07
streaming heartbeat query param on create of ws

himanshumeena:
this is the continuation of this pr: https://github.com/SaxoBank/openapi-clientlib-js/pull/416

earlier we are passing the sendHeartbeats query param in the update query method which get invoked when we reconnect or some failure happens but it doesn't add sendHeartbeats on initial ws socket creation for success case.

so moving the logic to create ws function which gets invoked in call cases.

### DIFF
--- a/src/openapi/streaming/connection/transport/websocket-transport.ts
+++ b/src/openapi/streaming/connection/transport/websocket-transport.ts
@@ -290,9 +290,13 @@ class WebsocketTransport implements StreamingTransportInterface {
 
     private createSocket() {
         try {
-            const url = this.normalizeWebSocketUrl(
+            let url = this.normalizeWebSocketUrl(
                 `${this.connectionUrl}${this.query}`,
             );
+
+            if (this.isWebsocketStreamingHeartBeatEnabled) {
+                url += '&sendHeartbeats=true';
+            }
 
             log.debug(LOG_AREA, 'Creating WebSocket connection', { url });
             const socket = new WebSocket(url);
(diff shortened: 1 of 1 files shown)


src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -283,10 +283,18 @@ class WebsocketTransport implements StreamingTransportInterface {
 
     private createSocket() {
         try {
-            const url = this.normalizeWebSocketUrl(
+            let url = this.normalizeWebSocketUrl(
                 `${this.connectionUrl}${this.query}`,
             );
 
+            if (this.isWebsocketStreamingHeartBeatEnabled) {
+                if (url.indexOf('?') === -1) {
```
lukeapage, line 291: the query always contains the authorization - below there are assumptions we can just add `&`

src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -283,10 +283,18 @@ class WebsocketTransport implements StreamingTransportInterface {
...
     private createSocket() {
         try {
-            const url = this.normalizeWebSocketUrl(
+            let url = this.normalizeWebSocketUrl(
                 `${this.connectionUrl}${this.query}`,
             );
 
+            if (this.isWebsocketStreamingHeartBeatEnabled) {
+                if (url.indexOf('?') === -1) {
+                    url += '?sendHeartbeats=true';
```
lukeapage, line 292: Please rebase on master

lukeapage (changes requested):

src/openapi/streaming/connection/transport/websocket-transport.ts
himanshumeena: yes that can be done
updated

src/openapi/streaming/connection/transport/websocket-transport.ts
himanshumeena: rebased with master

lukeapage (approved):

PR #467, opened 2021-07-13
Convert HTTP method names passed to fetch to uppercase

hiredgun:
This PR makes an http method name passed to fetch uppercase.

 See https://bugs.chromium.org/p/chromium/issues/detail?id=1228178

### DIFF
--- a/src/utils/fetch.ts
+++ b/src/utils/fetch.ts
@@ -204,10 +204,10 @@ export function convertFetchSuccess(
     return convertedPromise;
 }
 
-function getBody(method: HTTPMethodType, options?: Options): any {
+function getBody(method: Uppercase<HTTPMethodType>, options?: Options): any {
     // If PATCH without body occurs, create empty payload.
     // Reason: Some proxies and default configs for CDNs like Akamai have issues with accepting PATCH with content-length: 0.
-    if (method === 'patch' && !options?.body) {
+    if (method === 'PATCH' && !options?.body) {
         return {};
     }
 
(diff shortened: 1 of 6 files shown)


src/utils/fetch.ts
```diff
@@ -228,7 +228,7 @@ function localFetch<Response = any>(
     url: string,
     options?: Options,
 ): Promise<OAPIRequestResult<Response>> {
-    let method = httMethod.toLowerCase() as HTTPMethodType;
+    let method = httMethod.toUpperCase() as Uppercase<HTTPMethodType>;
```
robatwilliams, line 231: Why do we permit it to be called with the wrong case, and then fix it up here? I think it would be less mistake-prone if method names were always uppercase in this codebase, as is standard. One place I was expecting change is the `prepareTransportMethod()` calls in transport-base.

hiredgun: IIRC we switched to lowercase to simplify migration to classes and TS. For instance, some transports calls lower level transport methods dynamically (`transport.[method]`) so it was a trade-off between jumping between lower/uppercase in different places and having more unified naming and types. I think it also let us move some common logic to abstract classes and keeping the original interface. 
I can see the following advantages by doing it here:
- this wrapper can be used directly so in case someone uses the lib in non-ts project one is still covered and we take care of normalizing method names
- we are more compatible with the original fetch by allowing lowercase methods too
- the issues I mentioned above are still solved

robatwilliams (approved):

PR #503, opened 2021-08-18
fix ws taking long time to reconnect issue

himanshumeena:
- when client gets offline we do reconnect multiple time and each reconnect takes around 10 to 15 seconds.
- as soon as client gets online the last reconnect takes around 30 to 40 seconds before it finishes and the next reconnect call gets 101 instantly
- the idea behind this pr is to trigger a reconnect once client gets online so that we can stop the last reconnect call that takes 30 to 40 seconds so that client can get connected immediately for better user experience
- we are adding the 'online' event listener on window when websocket connection opens and removing the same when websocket connection destroys

### DIFF
--- a/src/openapi/streaming/connection/transport/websocket-transport.ts
+++ b/src/openapi/streaming/connection/transport/websocket-transport.ts
@@ -96,6 +96,8 @@ class WebsocketTransport implements StreamingTransportInterface {
     isWebsocketStreamingHeartBeatEnabled = false;
     inactivityFinderRunning: boolean;
     inactivityFinderNextUpdateTimeoutId: number | null = null;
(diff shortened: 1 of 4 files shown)


src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -107,6 +109,7 @@ class WebsocketTransport implements StreamingTransportInterface {
 
         this.failCallback = failCallback;
         this.inactivityFinderRunning = false;
+        this.reconnectOnceOnlineCallback = this.reconnect.bind(this);
```
lukeapage, line 112: do we want to be straight up reconnecting, without even checking if we are still connected?
What about signal-r core - I thought that had the same issue?

Issue #505, opened 2021-08-23
Probe messages are discarded


joshuarpearce:
As a user it isn't possible to use probe messages, or even trigger something when they arrive. This is due to the library picking them up for their "_" control message prefix but having no case for probe messages.

PR #506, opened 2021-08-23
Handle probe messages #505

joshuarpearce:
Working with probes the messages will always be discarded as they are picked up for their "_" prefix but not handled. This should handle them and allow the user to specifiy a callback to be triggered which will be provided with the message from the server.

### DIFF
--- a/src/openapi/streaming/control-messages.ts
+++ b/src/openapi/streaming/control-messages.ts
@@ -6,3 +6,4 @@ export const OPENAPI_CONTROL_MESSAGE_RESET_SUBSCRIPTIONS =
     '_resetsubscriptions';
 export const OPENAPI_CONTROL_MESSAGE_RECONNECT = '_reconnect';
 export const OPENAPI_CONTROL_MESSAGE_DISCONNECT = '_disconnect';
+export const OPENAPI_CONTROL_MESSAGE_PROBE = '_probe';
--- a/src/openapi/streaming/streaming.ts
+++ b/src/openapi/streaming/streaming.ts
@@ -19,6 +19,7 @@ import {
     OPENAPI_CONTROL_MESSAGE_CONNECTION_HEARTBEAT,
     OPENAPI_CONTROL_MESSAGE_RECONNECT,
     OPENAPI_CONTROL_MESSAGE_RESET_SUBSCRIPTIONS,
+    OPENAPI_CONTROL_MESSAGE_PROBE,
 } from './control-messages';
 
 const DEFAULT_CONNECT_RETRY_DELAY = 1000;
@@ -167,6 +168,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
     disposed = false;
     private heartBeatLog: Array<[number, ReadonlyArray<string>]> = [];
     shouldSubscribeBeforeStreamingSetup = false;
+    onProbe?: (message: types.ProbeControlMessage) => void;
 
     /**
      * @param transport - The transport to use for subscribing/unsubscribing.
@@ -179,11 +181,17 @@ class Streaming extends MicroEmitter<EmittedEvents> {
         baseUrl: string,
         authProvider: AuthProvider,
         options?: Partial<types.StreamingConfigurableOptions>,
+        callbacks?: {
+            onProbe: (message: types.ProbeControlMessage) => void
+        },
     ) {
         super();
         this.baseUrl = baseUrl;
         this.authProvider = authProvider;
         this.transport = transport;
+        if (callbacks?.onProbe) {
+            this.onProbe = callbacks.onProbe;
+        }
 
         this.setOptions({ ...DEFAULT_STREAMING_OPTIONS, ...options });
 
@@ -644,6 +652,10 @@ class Streaming extends MicroEmitter<EmittedEvents> {
                 // control mesage to keep streaming connection alive
                 break;
 
+            case OPENAPI_CONTROL_MESSAGE_PROBE:
+                this.handleControlMessageProbe(message);
+                break;
+
             default:
                 log.warn(LOG_AREA, 'Unrecognised control message', {
                     message,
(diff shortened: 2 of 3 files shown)


src/openapi/streaming/streaming.ts
```diff
@@ -167,6 +168,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
     disposed = false;
     private heartBeatLog: Array<[number, ReadonlyArray<string>]> = [];
     shouldSubscribeBeforeStreamingSetup = false;
+    onProbe?: (message: types.ProbeControlMessage) => void;
```
bobrippling, line 171: Let's make this private, matching the existing heartbeat stuff

src/openapi/streaming/streaming.ts
```diff
@@ -179,11 +181,17 @@ class Streaming extends MicroEmitter<EmittedEvents> {
         baseUrl: string,
         authProvider: AuthProvider,
         options?: Partial<types.StreamingConfigurableOptions>,
+        callbacks?: {
+            onProbe: (message: types.ProbeControlMessage) => void
```
bobrippling, line 185: If future callbacks are added, we won't want to force the user to specify them all - if we make this optional too, it'll save having to consider it when a later committer comes along

src/openapi/streaming/streaming.ts
```diff
@@ -689,6 +701,17 @@ class Streaming extends MicroEmitter<EmittedEvents> {
         }
     }
 
+    /**
+     * Handles probe control messages and calls given callbacks at creation time
+     * 
+     * @param message - the message from server
+     */
+    private handleControlMessageProbe(message: any) {
```
bobrippling, line 709: We can avoid `any` and type `message` as `ProbeControlMessage` here

src/openapi/streaming/streaming.ts
```diff
@@ -689,6 +701,17 @@ class Streaming extends MicroEmitter<EmittedEvents> {
         }
     }
 
+    /**
+     * Handles probe control messages and calls given callbacks at creation time
```
robatwilliams, line 705: I think part of this needs reordering and be singular - "calls callback given at"

src/openapi/streaming/streaming.ts
```diff
@@ -689,6 +701,17 @@ class Streaming extends MicroEmitter<EmittedEvents> {
         }
     }
 
+    /**
+     * Handles probe control messages and calls given callbacks at creation time
+     * 
+     * @param message - the message from server
```
robatwilliams, line 707: We have a lot of low/zero-value JSDoc in this project (so it looks like a convention which I guess you followed), but we tend to avoid adding more such ones

src/openapi/streaming/streaming.ts
robatwilliams: `ProbeControlMessage` instead of `any` ?

bobrippling (changes requested):

src/openapi/streaming/control-messages.ts
```diff
@@ -6,3 +6,4 @@ export const OPENAPI_CONTROL_MESSAGE_RESET_SUBSCRIPTIONS =
     '_resetsubscriptions';
 export const OPENAPI_CONTROL_MESSAGE_RECONNECT = '_reconnect';
 export const OPENAPI_CONTROL_MESSAGE_DISCONNECT = '_disconnect';
+export const OPENAPI_CONTROL_MESSAGE_PROBE = '_probe';
```
robatwilliams, line 9: All the other terms are pretty common/clear in streaming lingo. Could be worth a brief comment on what a probe is and what it's used for (bearing in mind this is a public repo)

src/openapi/streaming/streaming.ts
robatwilliams: This could live just as tidily on the options object - wdyt?

We have that pattern already in `subscription.ts` too.

src/openapi/streaming/streaming.ts
```diff
@@ -179,11 +181,17 @@ class Streaming extends MicroEmitter<EmittedEvents> {
...
         options?: Partial<types.StreamingConfigurableOptions>,
+        callbacks?: {
+            onProbe: (message: types.ProbeControlMessage) => void
+        },
     ) {
         super();
         this.baseUrl = baseUrl;
         this.authProvider = authProvider;
         this.transport = transport;
+        if (callbacks?.onProbe) {
```
robatwilliams, line 192: Are these 3 lines equivalent to `this.onProbe = callbacks?.onProbe;`

src/openapi/streaming/streaming.ts
joshuarpearce: Yep, makes sense

robatwilliams (commented):
Looks good, few small things for maintainability

src/openapi/streaming/streaming.ts
joshuarpearce: Yeah I'd confused services options with this so thought it wouldn't be suitable but after checking I agree with you. Changes made

robatwilliams (approved):

src/openapi/streaming/streaming.ts
```diff
@@ -184,6 +186,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
         this.baseUrl = baseUrl;
         this.authProvider = authProvider;
         this.transport = transport;
+        this.onProbe = options?.onProbe;
```
lukeapage, line 189: I expected this to do

this.trigger('probe')

rather than expect a new option for a callback.

lukeapage (changes requested):

src/openapi/streaming/streaming.ts
joshuarpearce: Agreed. Have made that change and tested it locally.

src/openapi/streaming/control-messages.ts
joshuarpearce: Done

lukeapage (approved):

src/openapi/streaming/types.ts
```diff
@@ -119,4 +128,8 @@ export interface StreamingConfigurableOptions {
      *  Flag to control whether we should subscribe before streaming setup during initial subscribe
      */
     shouldSubscribeBeforeStreamingSetup?: boolean;
+    /**
+     * Optional callback to handle probe messages
+     */
+    onProbe?: (message: ProbeControlMessage) => void;
```
bobrippling, line 134: This can go now too :)

PR #549, opened 2021-09-28
Make protobuf fallback to json

lukeapage:
This PR contains some fixes that came out of the bug with protobuf SchemaName fixed previously.

1. We should switch back to json if we get a protobuf error
2. We should be careful when we reset so we don't trigger the warning due to normal behaviour
3. We should log information on what is causing the 3 resets error message
4. If a update comes in before a snapshot and we process it during the initial snapshot and then get a parser error, we can swallow the subscribe part of the reset and so stay in the unsubscribed state

Testing:
- [x] I tested by rebasing this before the protobuf fix and verified that fallback to JSON works.

TODO:
- [ ] Unit tests covering old and new SchemaName behavior
- [ ] Unit tests covering fallback to JSON
- [ ] Unit tests covering errors processing the snapshot, so we do not unsubscribe and swallow the subscribe

### DIFF
--- a/src/openapi/streaming/streaming.ts
+++ b/src/openapi/streaming/streaming.ts
@@ -717,7 +717,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
     private resetSubscriptions(subscriptions: Subscription[]) {
         for (let i = 0; i < subscriptions.length; i++) {
             const subscription = subscriptions[i];
-            subscription.reset();
+            subscription.reset(true);
         }
     }
 
@@ -941,7 +941,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
         }
 
         this.connection.onOrphanFound();
-        subscription.reset();
+        subscription.reset(false);
     }
 
     private handleSubscriptionReadyForUnsubscribe(
@@ -1222,7 +1222,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
             const subscription = this.subscriptions[i];
             // Reset the subscription and mark it as not having a connection so its state becomes unsubscribed
             // next action if there is any i.e. subscribe will go in the queue until the connection is available again
-            subscription.reset();
+            subscription.unsubscribeAndSubscribe();
             subscription.onConnectionUnavailable();
         }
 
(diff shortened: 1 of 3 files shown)


src/openapi/streaming/streaming.ts
```diff
@@ -1254,7 +1254,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
             const subscription = this.subscriptions[i];
             // Reset the subscription and mark it as not having a connection so its state becomes unsubscribed
             // next action if there is any i.e. subscribe will go in the queue until the connection is available again
```
michax, line 1256: Is above comment stil valid?

src/openapi/streaming/subscription.ts
```diff
@@ -1020,8 +1044,17 @@ class Subscription {
...
+        this.checkIfPublisherDown(isServerInitiated);
+        this.unsubscribeAndSubscribe();
+    }
+
+    /**
+     * Does a unsubscribe and then schedules a subscribe for when it is finished unsubscribing
+     * Call this only if this is not a error scenario. Normally you should use reset
+     * So that we track the reset
+     */
+    unsubscribeAndSubscribe() {
```
michax, line 1057: Cannot comment on unchanged lines but below log might need adjusting as it's now not inclusive to reset.
```

log.error(
                    LOG_AREA,
                    'Reset was called but subscription is in an unknown state',
                );

```

src/openapi/streaming/streaming.ts
```diff
@@ -1283,7 +1283,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
             // disconnecting *should* shut down all subscriptions. We also delete all below.
             // So mark the subscription as not having a connection and reset it so its state becomes unsubscribed
             subscription.onConnectionUnavailable();
```
michax, line 1285: Do we still need to call onConnectionUnavailable if we dispose afterwards? Or is it kept to ensure we have logged info about specific disposal scenario?

michax (approved):

src/openapi/streaming/streaming.ts
lukeapage: updated

src/openapi/streaming/subscription.ts
lukeapage: updated

PR #558, opened 2021-10-05
Support cloud streaming control message format

mlister-scottlogic:
Add tests for the current message format.
Add tests and support for the new message format.
Log warnings when unsupported messages are found to find problems more quickly.

### DIFF
--- a/src/openapi/streaming/streaming.ts
+++ b/src/openapi/streaming/streaming.ts
@@ -604,7 +604,19 @@ class Streaming extends MicroEmitter<EmittedEvents> {
             return message.Heartbeats;
         }
 
-        if (message.Data && message.Data.length > 0) {
+        if (
+            message.Data &&
+            !Array.isArray(message.Data) &&
+            message.Data.Heartbeats
+        ) {
+            return message.Data.Heartbeats;
+        }
+
+        if (
+            message.Data &&
+            Array.isArray(message.Data) &&
+            message.Data.length > 0
+        ) {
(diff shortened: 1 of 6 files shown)


src/openapi/streaming/streaming.ts
```diff
@@ -604,10 +604,26 @@ class Streaming extends MicroEmitter<EmittedEvents> {
             return message.Heartbeats;
         }
 
-        if (message.Data && message.Data.length > 0) {
+        if (
+            message.Data &&
+            !Array.isArray(message.Data) &&
```
lukeapage, line 609: do we use this elsewhere in this library- I'm just aware it needs to be polyfilled for IE11. and in saxotrader, do we polyfill globally or via babel and will it therefore cover this?

I do prefer using Array.isArray if possible...

lukeapage (changes requested):

src/openapi/streaming/streaming.ts
mlister-scottlogic: Its used in a couple of places already - https://github.com/SaxoBank/openapi-clientlib-js/blob/aadfd1daaac1f9dc00609ef8c37af83793d75698/src/openapi/streaming/streaming.ts#L560 for example

mlister-scottlogic: Looks like its supported from IE9 too
https://caniuse.com/?search=Array.isArray

lukeapage (approved):

PR #571, opened 2021-10-13
Expose token refresh failure status

ikarami:
In case of a token refresh failure due to authorization error we expose the status with which the request failed. Based on that info different actions on the consumer side can be undertaken.

Adds test covering 407 status code (proxy authentication error).


src/openapi/authProvider.ts
```diff
@@ -99,7 +99,7 @@ const EVENT_TOKEN_REFRESH_FAILED = 'tokenRefreshFailed' as const;
 type EmittedEvents = {
     [EVENT_TOKEN_REFRESH]: () => void;
     [EVENT_TOKEN_RECEIVED]: (token?: string, refresh?: number) => void;
-    [EVENT_TOKEN_REFRESH_FAILED]: () => void;
+    [EVENT_TOKEN_REFRESH_FAILED]: (code?: number) => void;
```
hiredgun, line 102: isn't code always available given `isAuthenticationError` condition?

PR #596, opened 2021-11-08
Upgrade to 10.6.6 for subscription reset in case of missing message

prashantsharma31:
Upgrade to 10.6.6 for subscription reset in case of missing message


prashantsharma31:
as discussed its related changes are already merged it will be available in next version update so these changes not required

PR #600, opened 2021-11-09
Fix websocket transport error

lukeapage:


src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -526,57 +544,51 @@ class WebsocketTransport implements StreamingTransportInterface {
...
         const url = `${this.authorizeUrl}?contextId=${contextId}`;
 
-        this.authorizePromise = fetch('put', url, options)
-            .then((response) => {
+        fetch('put', url, options).then(
+            () => {
                 log.debug(LOG_AREA, 'Authorization completed', {
                     contextId,
                 });
-                return response;
```
lukeapage, line 534: The core of the issue here is that there is a single promise assigned to the class, but there can be multiple calls in progress. The start command might listen to `then` on a promise that does not complete

lukeapage: instead of getting a promise, you call a function to do something and there is a call to onAuthorised when it is done.

src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -526,57 +544,51 @@ class WebsocketTransport implements StreamingTransportInterface {
...
                 if (
                     this.authToken !== authToken ||
                     this.contextId !== contextId
                 ) {
-                    return Promise.reject(
-                        new Error(
-                            'Failed websocket-transport authorize with a different token',
-                        ),
-                    );
+                    return;
```
lukeapage, line 568: This was strange that this check is in the catch but not the then... it doesn't need a error because it isn't a error scenario

src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -526,57 +544,51 @@ class WebsocketTransport implements StreamingTransportInterface {
...
-                                }
-                                this.getAuthorizePromise(
-                                    contextId,
-                                    authToken,
-                                    true,
-                                ).then(resolve, reject);
-                            }, 300);
-                        },
-                    );
+                    setTimeout(() => {
```
lukeapage, line 573: keep retrying network errors every 300ms until a new token comes in.

src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -593,23 +605,40 @@ class WebsocketTransport implements StreamingTransportInterface {
...
 
-        const authorizePromise = this.getAuthorizePromise(
-            this.contextId,
-            this.authToken,
-        );
+        this.state = STATE_AWAITING_START;
 
-        authorizePromise.then(
-            () => {
-                this.startedCallback();
```
lukeapage, line 603: moved to onAuthorizedAwaitingStart

src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -669,34 +699,8 @@ class WebsocketTransport implements StreamingTransportInterface {
...
-                this.contextId,
-                authToken,
-                true,
-            );
-
-            authorizePromise.then(
-                () => {
-                    if (this.isReconnectPending) {
-                        this.isReconnectPending = false;
-                        this.reconnect(true);
```
lukeapage, line 692: moved to onAuthorizedAwaitingReconnect

src/openapi/streaming/connection/transport/websocket-transport.ts
```diff
@@ -649,6 +678,7 @@ class WebsocketTransport implements StreamingTransportInterface {
     ) {
         if (contextId !== this.contextId) {
             this.lastMessageId = null;
+            forceAuth = true;
```
lukeapage, line 681: I'm not sure this makes a difference. I think it makes sense, but I think if the context id changes then forceAuth will be true.
Except on start, but then I had to add a status guard below so that before start it doesn't request auth.

PR #616, opened 2021-12-01
Adds support for Word documents as a binary content type

kangelSL:
This PR adds support for Word documents (both .docx and .doc file types) as a binary content type
This is required because Word attachments can be uploaded in platforms but are corrupted on download


lukeapage:
what if we defaulted to binary instead of defaulting to text and having to have overrides for each binary type?

We could
1. Check if there are any text content types to whitelist
2. error in we fallback to text and log the content type
3. check the logs and if all is ok, change the default

kangelSL:
> what if we defaulted to binary instead of defaulting to text and having to have overrides for each binary type?
> 
> We could
> 
> 1. Check if there are any text content types to whitelist
> 2. error in we fallback to text and log the content type
> 3. check the logs and if all is ok, change the default

Hi @lukeapage - I have pushed a secondary change for the initial approach that could be taken - a textContentTypes variable, which can be checked at the same time as the multipart/mixed type. The default is still text, with a log on the `else` to check which variables are needed for the new text whitelist

I will look into checking which types hit this locally and advise how many text content types might be needed

lukeapage (approved):

PR #617, opened 2021-12-02
Add the url to the token rejected error message

lukeapage:


src/openapi/streaming/streaming.ts
```diff
@@ -308,7 +308,7 @@ class Streaming extends MicroEmitter<EmittedEvents> {
      * being rejected so we should get a new one.
      */
     private onUnauthorized() {
-        this.authProvider.tokenRejected();
+        this.authProvider.tokenRejected('streaming');
```
hiredgun, line 311: Shouldn't streaming URL be provided here so that it's compatible with what `tokenReject` expects? Maybe it's possible to somehow retrieve it from connection. Sims like it could provide some additional context during debugging.

changelog.md
```diff
@@ -1,3 +1,6 @@
+### v10.7.2
```
hiredgun, line 1: Please check if this will be the right version as Przemek plans to merge #609 (10.8.0)

src/openapi/streaming/streaming.ts
lukeapage: ok, it would be nice to correlate between different streaming services...

hiredgun (approved):

PR #662, opened 2022-02-03
Do not call message.cancelCallback() when no connection or connection is not in connected state

rafalszemraj:


src/openapi/streaming/connection/transport/signalr-core-transport.ts
```diff
@@ -370,13 +371,14 @@ class SignalrCoreTransport implements StreamingTransportInterface {
...
+                ( this.connection && this.connection.state === HubConnectionState.Connected)
                     // @ts-expect-error cancelCallback is no included in the IStreamResult but according to implementation it exists
-                    .cancelCallback()
-                    .then(sendCloseMessage)
-                    .then(() => this.connection && this.connection.stop())
-            );
+                    ? this.messageStream.cancelCallback()
+                    : Promise.resolve();
+            return closePromise
+                .then(sendCloseMessage)
```
lukeapage, line 380: won't send fail if we are not currently connected?
I wonder if we guard this too with checking the connection state?

rafalszemraj: good catch, connection.invoke calls sendMessage as well

PR #663, opened 2022-02-03
Fix fallback to protobuf

lukeapage:


src/openapi/streaming/subscription.ts
```diff
@@ -1042,21 +1042,33 @@ class Subscription {
...
+                    this.fallbackToJSON();
+                    // reset otherwise we do nothing and continue to get protobuf
+                    this.reset(false);
+                    return;
                 }
+
+                // when open api has upgraded this should be a warn, because we should always get the protobuf name
+                log.info(
+                    LOG_AREA,
+                    'Missing schema name, may cause protobuf errors',
```
ikarami, line 1065: why it says "missing schema name" if `this.SchemaName` has to be defined in order to reach this part?

ikarami (approved):

src/openapi/streaming/subscription.ts
lukeapage: I will make the error message clearer

PR #674, opened 2022-02-14
Add additional text content-type

lukeapage:


src/utils/fetch.spec.ts
```diff
@@ -360,6 +361,62 @@ describe('utils fetch', () => {
                 `);
     });
 
+    it('resolves if text type problem json', async () => {
```
hiredgun, line 364: With `test.each` we could also make sure that other types are handled properly e.g.
it.each`
    contentType
    {"application/problem+json; charset=utf-8"}
    {"application/x-javascript"}
`('resolves if text type is $contentType', ({contentType}) => {......

hiredgun (approved):

src/utils/fetch.spec.ts
lukeapage: yeah I didn't think it necessary to test every case - just one where we are splitting it

PR #677, opened 2022-02-16
Remove not required 'clearSubscriptionsInDispose' option

lukeapage:


changelog.md
```diff
@@ -1,3 +1,7 @@
+### v10.8.11
```
hiredgun, line 1: I guess this change may break TS so would just consider making it 10.9.0

lukeapage: ok

PR #748, opened 2022-06-10
Remove waiting for publisher state

lukeapage:
Another error is caused because the waiting for publisher state is set but we might be waiting for a patch to complete, so the state will get overridden with a subscribed state, taking us out of waiting for publisher. We may also do a reset when not needed, but again this is low impact.

Because the code bypasses waiting for publisher when we try and unsubscribe (try perform action), it might have items in the queue, but these are ignored (usually items in the queue must mean transitioning state, meaning we only add to the queue), it means that we process the unsubscribe, but we may have subscribe in the queue, so we do that after the unsubscribe. e.g.

1. waiting for publisher
2. call subscribe
3. call unsubscribe & remove
4. unsubscribe is processed
5. the subscribe is processed
6. the remove is processed

and the end result of that is that we are subscribed when we don't want to be subscribed.

I've fixed this by removing the state entirely. Instead.. if we are still in the same place as when the reset occurs, we do the reset, so it will just be x seconds later than expected.

But we never change the state and we never bypass a transitioning state by doing a action, which should avoid the bug.

### DIFF
--- a/src/openapi/streaming/subscription.ts
+++ b/src/openapi/streaming/subscription.ts
@@ -31,7 +31,6 @@ const stateFlags = {
     PATCH_REQUESTED: 16,
     REPLACE_REQUESTED: 32,
     READY_FOR_UNSUBSCRIBE_BY_TAG: 64,
-    PUBLISHER_DOWN: 128,
 } as const;
 
 export type SubscriptionState = typeof stateFlags[keyof typeof stateFlags];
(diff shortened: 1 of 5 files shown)


src/openapi/streaming/subscription.ts
```diff
@@ -290,15 +287,28 @@ class Subscription {
...
 
                 this.waitForPublisherToRespondTimer = window.setTimeout(() => {
                     this.waitForPublisherToRespondTimer = null;
-                    this.setState(this.STATE_UNSUBSCRIBED);
-                    this.tryPerformAction(ACTION_SUBSCRIBE);
+
+                    // only if nothing has changed - subscribed or mid-subscribed and the same reference id we got all the resets on
+                    if (
+                        referenceId === this.referenceId &&
+                        (this.currentState === this.STATE_SUBSCRIBED ||
```
prashantprabhakar, line 298: Suggestion:  maybe we could use `Array.includes` to check for currentState, like
```[this.STATE_SUBSCRIBED, this.STATE_SUBSCRIBE_REQUESTED, this.STATE_PATCH_REQUESTED, this.STATE_REPLACE_REQUESTED].includes(this.currentState)```

lukeapage: thanks, it is a little bit more readable.

prashantprabhakar (approved):

PR #786, opened 2022-07-09
Add subscription as argument on update and error callbacks

lukeapage:


src/openapi/streaming/subscription.ts
```diff
@@ -1159,7 +1163,7 @@ class Subscription {
         }
 
         // Serialization of Snapshot is not yet supported.
-        this.onUpdate?.(response.Snapshot, this.UPDATE_TYPE_SNAPSHOT);
+        this.onUpdate?.(response.Snapshot, this.UPDATE_TYPE_SNAPSHOT, this);
```
ikarami, line 1166: nitpick - why those types are repeated within the class itself even though they are already defined outside as:

const updateTypes = {
    UPDATE_TYPE_SNAPSHOT: 1,
    UPDATE_TYPE_DELTA: 2,
} as const;

src/openapi/streaming/subscription.ts
```diff
@@ -45,24 +45,28 @@ export interface StreamingOptions {
      * @param data - data received
      * @param updateType - either be subscription.UPDATE_TYPE_DELTA or subscription.UPDATE_TYPE_SNAPSHOT
      */
-    onUpdate?: (data: unknown, updateType: SubscriptionUpdateTypes) => void;
+    onUpdate?: (
+        data: unknown,
+        updateType: SubscriptionUpdateTypes,
+        subscription: Subscription,
```
ikarami, line 51: please update JSDoc for every function that now accepts subscription as a parameter

ikarami (approved):
Two minor issues, but LGTM

src/openapi/streaming/subscription.ts
lukeapage: I don't know but I don't have much of a preference myself. I guess this way it is consistent with how it is checked (consumers use subscription.UPDATE_TYPE_SNAPSHOT). I agree its a bit messy and would be better exposed separately, but  I think its a bigger standalone task.

src/openapi/streaming/subscription.ts
lukeapage: Fixed

PR #885, opened 2022-10-05
Fix orphan issue

lukeapage:
Came from investigating why we have to many orphan logs. In conseqence this shouldn't have much of a end user effect, except it might be longer before the subscription is reset (because the orphan finder won't reset it)

### DIFF
--- a/src/openapi/streaming/subscription.ts
+++ b/src/openapi/streaming/subscription.ts
@@ -139,7 +139,7 @@ export function __forTestingOnlyResetReferenceId() {
 
 const DEFAULT_REFRESH_RATE_MS = 1000;
 const MIN_REFRESH_RATE_MS = 100;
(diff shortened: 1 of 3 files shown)


rafalszemraj (approved):

src/openapi/streaming/subscription.ts
```diff
@@ -303,14 +303,19 @@ class Subscription {
                 this.resetTimeStamps[2] - this.resetTimeStamps[0] <
                     MIN_WAIT_FOR_PUBLISHER_TO_RESPOND_MS
             ) {
-                // 3 reset within 1 minute so wait for 1 minute for publisher to respond
+                // 3 reset within 1 minute so wait for 30 seconds for publisher to respond
```
rafalszemraj, line 306: And here as well

rafalszemraj (approved):

rafalszemraj (approved):

PR #907, opened 2022-10-26
Retry initial connection under experimental flag

lukeapage:
Firstly, some streaming connections are falling back to long polling, so I added a option to allow the initial connect to retry before falling back to long polling.

Then I looked at why the initial connection being blocked did not result in detection of lack of a network connection - it is because no reconnecting event is being sent. Then I found that we were not subscribing if the subscribe request came in after streaming had failed, because by then we set the option to subscribe before setup to false, so I changed that around to take a different approach.


rafalszemraj (approved):

rafalszemraj (approved):
Approved

